Issue #9, opened 2026-04-12
Original firmware


michkovitalik:
Hi. Is there a way to revert to the original firmware? People have been asking me on the 4pda forum.

michkovitalik:
They also want to become testers, but they are afraid that if something doesn't work out, they won't be able to return to stock.1.2.0

Komzpa:
I managed to make a way to go to original firmware while keeping custom bootloader so you don't need to take the device apart many times: https://github.com/DavidClawson/OpenScope-2C53T/pull/16 - go to openscope for fun, go to stock when need the things that don't work yet on openscope

DavidClawson:
@michkovitalik — you asked this on 12 April, on behalf of the 4pda forum, and the reason you gave was the right one:

> They also want to become testers, but they are afraid that if something doesn't work out, they won't be able to return to stock 1.2.0

**That fear is now addressed. #16 is merged** (`e91c705`), so you can flash between stock V1.2.0 and OpenScope from the PC, as often as you like, with the case closed.

## What you get

```bash
python3 scripts/switch_firmware.py stock --flash    # back to stock V1.2.0
python3 scripts/switch_firmware.py openscope        # back to OpenScope
```

The important properties, all of which @Komzpa verified on his own 2C53T before this merged:

- **The custom bootloader survives both directions.** This is the part that makes it repeatable — no BOOT0 jumper, no opening the case, no ROM DFU after the first setup.
- **Your stock settings survive the round trip byte-for-byte.** He saved the stock settings page, went to OpenScope and back, and compared: exact match. Language, calibration, preferences — the device comes back as you left it, not factory-reset.
- **Every block is CRC-verified on write**, and there's a readback check of the app, the stock settings page and the recovery bootloader afterward.

He ran the full cycle both ways and finished by restoring his own unit to stock, so the "can I get back?" question was answered by actually getting back.

## Two honest caveats

**This is what makes testing safe, not what makes the firmware ready.** Please pass that on to 4pda along with the good news. **The oscilloscope does not capture yet** — that's the [#18](https://github.com/DavidClawson/OpenScope-2C53T/issues/18) FPGA problem, which is where nearly all the project's effort goes. The multimeter works but its calibration factor is currently hardcoded to my bench unit, so readings should be checked against a known-good meter. Anyone who wants a working scope today should stay on stock. Anyone who wants to *help* now has a safety net.

[...]

PR #14, opened 2026-06-04
docs: document ROM DFU recovery path

Komzpa:
## Summary
- document the bootloader-only outcome after a ROM DFU app write/leave failure
- add the no-programmer `unprotect:force` recovery path for AT32 ROM DFU `SET_ADDRESS` failures
- note the Linux `/bin/sh` vs bash `option_bytes48.bin` size check

## Verification
- `git diff --check README.md docs/dfu_mode_guide.md`
- validated against a real first-time flash/recovery session on FNIRSI 2C53T

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,11 @@ The short version:
    ```bash
    make flash-all
    ```
-7. Remove the BOOT0 jumper, pinhole reset, close the case — you won't need to open it again
+   If the second `dfu-util` command reports `File downloaded successfully` but
+   fails while submitting `leave`, do not assume the application has booted.
+   Remove the BOOT0 jumper, reset into the USB HID bootloader, and run
+   `make flash` to install the application through the bootloader.
+7. Remove the BOOT0 jumper, pinhole reset, close the case — you won't need to open it again once the application boots
 
 ### Normal Development Cycle (case closed)
 
--- a/README.md
+++ b/README.md
@@ -126,10 +126,10 @@ The short version:
    ```bash
    make flash-all
    ```
-   If the second `dfu-util` command reports `File downloaded successfully` but
-   fails while submitting `leave`, do not assume the application has booted.
-   Remove the BOOT0 jumper, reset into the USB HID bootloader, and run
-   `make flash` to install the application through the bootloader.
+   If the application write finishes but the device does not come up running the
+   app, do not assume it booted. Remove the BOOT0 jumper, reset into the USB HID
+   bootloader, and run `make flash` to install the application through the
+   bootloader.
 7. Remove the BOOT0 jumper, pinhole reset, close the case — you won't need to open it again once the application boots
 
 ### Normal Development Cycle (case closed)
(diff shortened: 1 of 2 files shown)


DavidClawson:
@Komzpa — on the merge track too. The ROM DFU recovery path and the `unprotect:force` route for `SET_ADDRESS` failures are real gaps, and "validated against a real first-time flash/recovery session" is exactly the bar I want for the DFU guide. The `/bin/sh` vs bash `option_bytes48.bin` size-check note is a nice catch — that's bitten Linux users. Thank you.

DavidClawson:
@Komzpa — this sat for 68 days after I said "on the merge track". Sorry. **Merging now.**

I pushed one maintainer commit on top of yours (`19d6ee4`) rather than asking you to rebase, because **every change I made is there because the tree moved under this PR while it was open — not because anything here was wrong when you wrote it.** Your commit is untouched. Details so nothing lands in the docs without you having seen it:

**1. The app write is no longer `dfu-util ...:leave`.** Your text keyed the advice off "the second `dfu-util` command … fails while submitting `leave`". Since the switcher work, `make flash-all` writes the application with `at32_dfuse_write.py --leave 0x08004000`, so that error message no longer appears. I rewrote it around the observable symptom instead — the device comes up on `BOOTLOADER MODE` rather than the app — and added that `leave status=... state=...` and `device detached after leave: ...` are the **normal** output of that step, since both look alarming and neither is a failure.

The advice itself was right and is unchanged: it means the one-time ROM DFU work succeeded, and `make flash` through the bootloader finishes the job. That's a genuinely useful thing to have written down; it's the point at which a first-time user is most likely to think they've bricked the device.

**2. The `/bin/sh` note — you fixed this bug yourself, three weeks after writing the note.** I checked it on Linux before touching it, because I wanted to be sure the note was obsolete rather than just unfamiliar:

```
$ /bin/sh -c "printf '\xA5\x5A...'" | wc -c     # dash, the old hex form
192
$ /bin/sh -c "printf '\245\132...'" | wc -c     # dash, the octal form
48
```

**Your report was exactly right** — under `dash` the old `printf '\xHH'` produced 192 bytes, not 48, and that would have failed the option-byte write with "address out of range". But `e6c2696` in #16 converted the recipe to octal escapes, which are portable, so `make SHELL=/bin/bash` now sends people chasing a problem that no longer exists. I kept the byte-count check and the history of what used to go wrong, and dropped the workaround.

**3. `unprotect:force` needed a stronger warning after #16.** It said "This erases the application flash". That's now an understatement: on a device set up with the switcher it also destroys the **archived stock V1.2.0 image and the high recovery bootloader** — the stock-restore path #9 was closed on yesterday. The entry now says the recovery is whole-flash, that first-time setup has to be redone, and that a stock image has to be re-archived before the switcher works again.

**Everything else merged as written**, including the `SET_ADDRESS` recovery itself, the "only close the case after the application has booted" instruction, and the `BOOTLOADER MODE` troubleshooting entry. The `unprotect:force` route is the part I'm most glad to have documented — it's the difference between "no external programmer required" and "buy a programmer", and I had no idea it worked.

"Validated against a real first-time flash/recovery session" is the bar I want for this guide, and you hit it. Thanks — and sorry again for the wait.

Issue #1, opened 2026-04-04
Test point Boot0


michkovitalik:
Can you tell me where the boot0 test point is? It's not signed for me.

baraa1936:
i can confirm that you can do it using the usb without the battery connected

PR #16, opened 2026-06-06
boot: add PC-driven stock/OpenScope firmware switcher

Komzpa:
## Summary

- add a high-recovery HID layout so the device keeps a resident recovery path at `0x080F0000` while the stock APP runs at its native linked address
- add a stock launcher and `scripts/switch_firmware.py` commands for PC-driven switching between stock firmware and OpenScope
- preserve the stock saved-settings page at `0x08006000..0x08007000` when switching firmware, including user language and startup mode
- fix the boot-counter magic from `0xB007F000` to `0xB0070000`, making the existing 3-strike safe-mode counter actually match under `BOOT_COUNTER_MASK`
- recover the stock language setting from live read-only dumps: English is `0x02`, Chinese is `0x01` at `0x0800602C`
- export recovered stock settings as an OpenScope config overlay for the fields that are safe to share now: startup mode and language
- patch the stock battery-only handoff so PC9 power-hold stays latched while stock starts
- add an optional fast FNIRSI splash patch; the default stock-user image skips the cosmetic 256-step logo delay while leaving hardware init intact
- add guarded preflight, CRC/readback, and unit tests so stock APP bytes are not accidentally flashed as an app-slot image and low-flash writes require high-recovery unlock

## Review Fixes

Current head `7eda07d6d209fea94c71169495e9fe9a10fabb4e` addresses the July 28 review blockers:

- restored tracked `firmware/stage0/` sources and made `firmware/high_layout_updater` build its stage0 and high-bootloader prerequisites from a clean checkout
- restored `scripts/update_bootloader_via_hid.py`, matching the documented bootloader-updater entry point
- changed `flash-dfu` and `flash-all` to use the preserve-aware AT32 DFU writer for the app image instead of raw app-slot `dfu-util -D firmware.bin`
- made `jump_to_app()` return on invalid vectors again, so the bootloader can stay recoverable over USB instead of hanging
- hardened HID preserve logic at the real 2KB erase-sector level: ranges must be sector-aligned and a preserve selection may not cover only one 1KB half-sector
- updated README, CLAUDE notes, app-side DFU comments, and watchdog fault text from POWER-alone recovery to POWER+PRM

## Stock Layout

The archived V1.2.0 stock APP is linked for `0x08007000`:

```text
APP sha256: a17c5c35c97bb898f15672a1747bc1041d8ed507c16999ddba0d1e4e2ec0c760
APP size:   751232 bytes
vector[0]:  0x20036F90
vector[1]:  0x08007311
```

The generated stock-user image layout is:

```text
0x08000000: low reset vector / stage0
0x08007000: stock APP bytes with runtime safety patches
0x080E0000: stock launcher
0x080F0000: high HID recovery bootloader
```

For the archived APP, the stock settings preserve floor is `0x080BE800` (`0x08007000` plus the padded stock APP size). Low padding below that floor is erased/programmed normally; only whole blank erase sectors in the post-stock all-`0xFF` hole and the stock saved-settings page are preserved.

This layout replaces the original `0x08000000` factory path. After installing it, reversibility comes from keeping the stock APP resident plus the high HID recovery bootloader, not from preserving the factory MENU+Power upgrade mode as the active safety net.

## Settings Sharing

Stock saves a small internal-flash settings blob at `0x08006000`. This PR keeps that page intact when switching firmware and now decodes the recovered fields with:

```bash
python3 scripts/hid_mem_read.py 0x08006000 0x1000 --stock-settings
python3 scripts/stock_settings.py inspect settings.bin
python3 scripts/stock_settings.py export-openscope-overlay settings.bin
```

The currently recovered shared fields are:

- startup mode: the saved-mode low byte identifies stock multimeter startup and maps to OpenScope `CONFIG_MODE_MULTIMETER`
- language: live English -> Chinese -> English readback confirmed `0x0800602C` toggles `0x02 -> 0x01 -> 0x02`, mapping to OpenScope English/Chinese config values

Unrecovered values remain explicit `unknown` rather than guessed.

## Fast Stock Splash

The default stock-user builder now patches the stock FNIRSI logo loop from initial step `0` to `255`:

```text
stock APP offset: 0x00024506
stock address:    0x0802B506
original bytes:   00 20
patched bytes:    ff 20
estimated saving: 768 ms
```

This skips the cosmetic 256-step logo delay only. The builder report keeps `hardware_init_skipped_by_fast_splash=false`; LCD, SPI/FPGA, and post-config waits remain in place. Use `--keep-splash` for an A/B image with vanilla splash timing.

## Usage

Build stock-user image without flashing:

```bash
python3 scripts/switch_firmware.py stock
```

Flash and run stock from high HID recovery, preserving stock settings:

```bash
python3 scripts/switch_firmware.py stock --flash
```

Flash stock but stay in HID recovery for readback checks:

```bash
python3 scripts/switch_firmware.py stock --flash --no-jump
```

Flash and run OpenScope from HID recovery:

```bash
python3 scripts/switch_firmware.py openscope --build
```

## Current Limits

A sealed, fully stock device still needs a first path into our bootloader or ROM DFU. The branch adds `hid_enter_rom_dfu.py` and keeps ROM-DFU entry available from the custom HID bootloader, but it does not yet add a stock-install guest app that jumps from factory IAP into ROM DFU. That guest bridge looks plausible, but it is not part of this PR.

## Validation

Current head: `7eda07d6d209fea94c71169495e9fe9a10fabb4e`

Fresh local validation on July 28, 2026:

```bash
python3 -m py_compile scripts/*.py
git diff --check
for t in scripts/test_*.py; do python3 "$t" || exit $?; done
make -C firmware/stage0
make -C firmware/bootloader high
make -C firmware/high_layout_updater
make -C firmware
make -C firmware/bootloader
make -C firmware/bootloader_updater
```

Results:

```text
firmware/stage0:                 1004 B / 2048 B
firmware/bootloader high:        12056 B / 65536 B
firmware/high_layout_updater:    26900 B
firmware app:                    509492 B
firmware/bootloader:             11608 B / 16384 B
firmware/bootloader_updater:     25156 B
Python test files:               all passed
```

The focused validation exited 0 locally. The firmware build still emits existing warnings in unrelated files (`scope_ui.c`, `dac_output.c`, and SDK HID class code), but no build failure.

Live device readback on the rebased branch lineage:

- stock settings were preserved across stock-user reflashes and decoded as multimeter startup plus English language
- language was confirmed by read-only stock settings dumps across English -> Chinese -> English
- before flashing fast splash, live readback at `0x0802B506` showed vanilla bytes `00 20`
- after flashing the default stock-user image, readback at `0x0802B506` showed fast-splash bytes `ff 20`
- PC9 power-hold safety patch read back at `0x0802AAB6` as `00 bf`
- stock was launched from HID recovery with `CMD_RUN_ADDR 0x08000000`

This PR addresses the recovery confidence problem behind DavidClawson/OpenScope-2C53T#9: users can return to stock from a closed case through the resident high-recovery HID path, while stock settings survive firmware switching.


DavidClawson:
@Komzpa — this is a serious piece of work, thank you. The dual-boot switcher is exactly what unblocks adoption: issue #9, and most of the 4pda hesitation michkovitalik has been relaying, comes down to *"I'm afraid I can't get back to stock"* — and a PC-driven `switch_firmware.py stock --flash` / `openscope --build` is the answer.

A few things I appreciate from reading through it:
- Keeping the archived stock APP byte-exact at its linked `0x08007000` and verifying by CRC instead of relinking it — that's the safe call.
- High HID recovery resident at `0x080F0000`, outside the stock range, so there's always a way back.
- The preflight guards (never flashing stock APP bytes as an app-slot image; requiring high-recovery unlock for low-flash writes) and preserving the stock settings pages by skipping the all-`0xFF` blocks above the settings hole. That defensive plumbing is what makes me comfortable pointing others at it.

Give me a bit to review properly — it replaces the bootloader and touches the recovery path, so I want to walk it carefully and bench it on a dev unit before merging (I keep one unit as a pristine read-protected reference and won't risk that one on a bootloader swap). But the design is right, and I'm glad you built it. I've also been doing SWD/option-byte work this week — I'll fold those findings into the review so we stay consistent.

Komzpa:
@DavidClawson I fixed a couple things found when using it - before it can only be switched on with the stock+bootloader when charging (bootloader passed control to firmware but it immediately went back to sleep instead of poweron unless it needs to display charging animation), now it works better. Also added storing stock settings (so it can be migrated over to openscope and/or back, or just not lost on reflash), and made the boot splash on stock faster - it now only is there for initialization, not for marketing.

DavidClawson:
@Komzpa — I pulled the branch and put it through a full build + test pass on a Linux bench box (Ubuntu 24.04, `arm-none-eabi-gcc` 13.2.1). Thank you for coming back to this with real device time behind it; the three fixes you describe are exactly the ones that separate "prototype" from "something a stranger can use."

## What I verified

| Target | Result |
|---|---|
| `firmware/bootloader` | ✅ 12,024 B / 16 KB (73%) |
| `firmware/bootloader` `make high` | ✅ 12,468 B @ `0x080F0000` |
| `firmware/stock_dispatcher` | ✅ 1,256 B |
| `firmware/bootloader_updater` | ✅ 25,784 B |
| `firmware/high_layout_updater` | ❌ blocked — see (1) |
| app, `make` | ✅ 504,840 B |
| app, `make guest` | ✅ 492,904 B |
| Python test suite | ✅ **7 of 8 files pass clean**, 8th passes once build artifacts exist |

I checked the flash layout against `objdump` rather than taking it on faith: `.isr_vector` lands at `0x08004000`, `.text` at `0x08007000`, and `0x08006000–0x08006FFF` is a genuine 0xFF hole in the image. The `ASSERT` guarding vector-table overflow into that gap is a good touch — it turns a silent corruption into a link error.

## Three things I want to credit specifically

**The `BOOT_COUNTER_MAGIC` change is a real bug fix, and I don't think you flagged it as one.** With `BOOT_COUNTER_MAGIC = 0xB007F000` and `BOOT_COUNTER_MASK = 0xFFFF0000`, the masked read `(boot_val & 0xFFFF0000)` always has its low 16 bits clear, so it could never compare equal to a constant carrying `F000` down there. The magic check failed on every boot, `fail_count` never accumulated, and **the 3-strike safe-mode system has been silently dead since it was written**. `0xB0070000` makes it actually work. Your comment in `dfu_boot.c` shows you traced the whole flow. That's a latent-safety-net bug found by reading, and it deserves calling out in the PR description.

**The privilege layering on the new IAP commands is well done.** Low-flash write, run-at-address, and read-memory are all behind `#ifdef HIGH_IAP_ALLOW_LOW_FLASH`, which the Makefile defines only for the `high` target — plus a magic word on the unlock, plus a redundant `BOOTLOADER_BASE_ADDRESS >= FLASH_APP_END_ADDRESS` check that independently fails on a `0x08000000` build. Three layers, and the bootloader ordinary users run gets none of the new attack surface. I went looking for a hole here and didn't find one.

**The portability fixes are quietly valuable.** The `nano.specs` probe and the octal `printf` escapes fix exactly the class of breakage reported in #19 on msys2 and Raspberry Pi. Those will save someone an evening.

## Findings

Two blockers, four smaller ones.

1. **`firmware/stage0/` isn't in the branch.** `high_layout_updater/Makefile:18` needs `../stage0/build/stage0.bin`, and `build_high_layout_payload.py` enforces a 2 KB cap on it, but the directory isn't in this PR or in `main`. I'd guess it's untracked locally. That target can't build for anyone but you right now.
2. **`scripts/update_bootloader_via_hid.py` doesn't exist**, but `bootloader_updater/README.md` documents it as the entry point for both the dry run and the live update. Same cause, probably.
3. **`flash-dfu` and `flash-all` bypass the settings preservation.** `make flash` correctly passes `--preserve-blank-blocks --preserve-blank-blocks-range 0x08006000:0x08007000`, but both DFU targets do a raw `dfu-util -s 0x08004000 -D firmware.bin`, and that image now carries the settings page as 4 KB of 0xFF — so it silently erases what the PR is designed to preserve. Defensible for `flash-all` (first-time setup), but `flash-dfu` is a fallback people reach for later.
4. **`jump_to_app()` no longer returns on an invalid image.** Both validation failures changed from `return` to `while (1) { __NOP(); }`. Previously an invalid app fell through to the caller and landed in HID IAP — recoverable over USB. Now it hangs. In the recovery bootloader specifically, I'd argue that's the one place the failure mode should stay recoverable.
5. **Latent sector-alignment hazard in the block-preserve logic.** Blocks are 1 KB, but `iap_erase_sector()` only fires on 2 KB-aligned addresses. If a preserved blank block ever lands on the aligned first half of a 2 KB sector while the second half gets written, that write goes to unerased flash. Your current `0x08006000:0x08007000` range is exactly two whole sectors, so it's safe today — but `--preserve-blank-blocks-from` could put a user outside that. An assert that preserve ranges are sector-aligned would lock the property in.
6. **POWER-alone bootloader entry is gone** (POWER+PRM now). Your reasoning is sound and well documented in the comment — POWER-alone is the battery power-on gesture and was dropping battery-only stock boots into recovery. But our README and `CLAUDE.md` still document "hold POWER at boot = force bootloader" as the recovery gesture, so that documentation needs to change in the same PR or users will reach for a chord that no longer exists.

## The design question I'd most like your view on

One of our standing goals is that a new user with a sealed, stock device can install OpenScope and get back to stock **without ever opening the case** — no BOOT0 jumper.

Reading this branch, I think you've built two of the three pieces for that, maybe without framing them that way:

- `jump_to_rom_dfu()` + `hid_enter_rom_dfu.py` means a device can be commanded into ROM DFU over USB, which removes the BOOT0 jumper — and it's deliberately *not* behind `HIGH_IAP_ALLOW_LOW_FLASH`, so the ordinary bootloader has it too.
- `bootloader_updater` means bootloader updates no longer require DFU at all.

The gap I can see is that both assume our HID bootloader is *already installed*. There's no ROM-DFU jump in `firmware/src/`, and the factory IAP only writes the `0x08007000` app slot — so a device that's still fully stock has no software route into ROM DFU, and first install still needs the jumper once.

Would dropping the same `jump_to_rom_dfu()` into the app (guest-slot) build work? If so the chain becomes: stock → MENU+Power → drop a guest image → app jumps to ROM DFU → `dfu-util` writes the bootloader and app, case never opened.

(On the related option-byte question: our setup docs have long said EOPB0=0xFE needs ROM DFU, but re-reading our own decompilation of stock — `spi_flash_init_controller` at `0x0802cf7c` — stock writes EOPB0=0xFE itself on first boot behind an `if already 0xFE, skip` guard. So any device that has ever booted stock should already have 224KB SRAM, and that's probably *not* a second reason to open the case. Static analysis only, not yet confirmed on a device that never had ROM DFU applied — if anyone reading this has a stock unit and is willing to check, I'd like to close that loop.)

Related and worth being explicit about, since it bit @michkovitalik in #17: this layout replaces `0x08000000`, so the factory MENU+Power upgrade mode goes away. Its reversibility comes from keeping the stock APP resident instead. That's a legitimate design, but it's a different guarantee from "factory update mode still works," and I think the README should say so plainly so nobody is surprised about which safety net they still have.

## Where I'd like to take this

This is the closest the project has come to solving the stock-restore blocker, and I want to see it land. Two things stand between here and that:

- If you can push `firmware/stage0/` and `scripts/update_bootloader_via_hid.py`, the branch becomes reproducible end-to-end and I can test the full path rather than the pieces.
- Then a hardware pass on my side. Fair warning on timing: our FPGA config-entry investigation currently depends on the factory IAP MSC channel to flash patched stock images (our app-side USB CDC is down), and this PR replaces that channel — so I'll want to run it on a second unit or once that work clears, rather than risk the bench box mid-experiment.

None of that is a criticism of the branch; it's scheduling. The code is in much better shape than the last round, and the fact that you found the battery power-on bug by actually living with it is worth more than any amount of review I can do from a diff.

DavidClawson:
@Komzpa — re-reviewed at `7eda07d`, sorry for the slow turnaround. Built and ran everything rather than reading the diff, on Ubuntu 24.04 with `arm-none-eabi-gcc` 13.2.1.

## Verified

| Finding | Status |
|---|---|
| ① `firmware/stage0/` missing, target unbuildable | ✅ **builds** — `stage0.elf`, 1076 B text. `high_layout_updater` builds too (15,700 B) |
| ② `scripts/update_bootloader_via_hid.py` missing | ✅ present |
| ③ `flash-dfu`/`flash-all` bypassing settings preservation | ✅ covered in `at32_dfuse_write.py` + `hid_flash.py` |
| ④ `jump_to_app()` `while(1)` hang | ✅ removed; POWER-alone recovery replaced by the bootloader's POWER+PRM check |
| ⑤ 1 KB-block vs 2 KB-erase alignment | ✅ host-side — see note below |
| ⑥ POWER+PRM entry undocumented | ✅ README |

All four test scripts pass: `test_flash_preflight.py`, `test_flash_switcher.py`, `test_bootloader_power_entry.py`, `test_bootloader_updater.py`.

On ⑤ specifically — `validate_preserved_sectors()` goes further than the finding asked. It walks the image sector by sector and rejects a *partial* preserve, which catches the case I under-described: a range can be perfectly aligned and still preserve only one of the two blocks, because qualification also depends on content being blank. Alignment alone wouldn't have caught that. Both validators run before USB is opened or anything is erased, and `parse_preserve_range` enforces it at argparse level so the raw-DFU path is covered too.

## One thing still open

`iap_erase_sector()` (`firmware/bootloader/src/hid_iap_user.c:159`) still silently no-ops when the address isn't 2 KB-aligned:

```c
else if (iap_info.sector_size == SECTOR_SIZE_2K)
{
  if ((address & (SECTOR_SIZE_2K - 1)) == 0)
    flash_sector_erase(address);
}
```

No `else`, and `iap_address()` ACKs regardless — so a misaligned address reports success and later writes land in unerased flash. The invariant is enforced **host-tool-side only**. Anything driving the HID protocol without the new `hid_flash.py` (an older copy, a hand-rolled script, direct protocol poking) can still trip the original hazard.

Defensible as-is — the bootloader is permanent and has ~4 KB headroom in 16 KB — but the PR reads as though the hazard is closed outright, and it's conventional rather than structural. A NACK on misalignment would make it structural. Your call whether that belongs here or in a follow-up; I don't think it should block the merge.

## Note on the review process

Two of the six findings were fixed in files whose names pointed at a *different* finding, and I initially mis-scored ⑤ as unaddressed because I inferred coverage from the changed-file list instead of reading the diff. That was my error, not a gap in your work.

Thanks for turning six findings into a working `stage0` in four hours. The dual-boot switcher is the thing that unblocks adoption — "can I get back to stock?" is the single most common hesitation we hear.

Komzpa:
@DavidClawson PTAL

DavidClawson:
@Komzpa — reviewed `11e4f76`. I said the device-side NACK didn't have to block the merge and you had it pushed twelve minutes later, so: same setup as before, Ubuntu 24.04 with `arm-none-eabi-gcc` 13.2.1, built and run rather than read.

## Verified

| Check | Result |
|---|---|
| Bootloader builds | ✅ **12,096 / 16,384 B (73%)** — +72 B over `7eda07d`, 4,288 B headroom |
| Other targets still build | ✅ `stage0` 1,076/2,048 (52%), `high_layout_updater` 12,540/65,536, `bootloader_updater` 25,856 B, `stock_dispatcher` 1,256/65,536 |
| Test suites | ✅ **49 pass** — `test_flash_preflight` 21, `test_flash_switcher` 13, `test_bootloader_power_entry` 2, `test_bootloader_updater` 13. The 5 skips are all `stock archive image is not present` |
| New warnings | ✅ none from project source |
| Signature change fallout | ✅ `iap_erase_sector` has exactly one caller (`hid_iap_user.c:268`) and it checks the return; no prototype in any header, so `void` → `iap_result_type` is fully contained |

The finding is closed the way I hoped: on an invalid address `iap_info.iap_address` is never set, `iap_info.state` never advances to `IAP_STS_ADDR`, and the host gets a NACK. The invariant is device-side now rather than a convention the host tooling upholds.

**The `return 0` in `iap_erase_address_valid()` closes something I didn't report.** The old code, given a `sector_size` it didn't recognise, matched none of the three branches — so it skipped the erase, ACKed, and advanced the state. Same failure as the misalignment case but quieter, because there's no wrong-looking address to notice afterwards. Default-deny covers it.

**I did go looking for the regression that default-deny could introduce**, since it turns a previously-silent case into a hard refusal: if `sector_size` were ever anything but 1K or 2K, flashing would now NACK outright. It can't be. `iap_init()` sets it to 1K below 256 KB of flash and 2K otherwise — never 4K, never 0 — and runs from both `iap_idle()` and `iap_start()` ahead of any ADDR command, so on the AT32F403A's 1 MB it is always 2K. And in the pathological case where an ADDR arrives first, `iap_info` is a zero-init global, `sector_size` is 0, and the new code NACKs where the old one would ACK and then write into unerased flash. Strictly safer on every path I can construct.

## Three notes, none blocking

1. **The return check at line 268 is unreachable-else.** `iap_erase_address_valid(address)` already returned true at 262, and that is the only condition under which `iap_erase_sector()` returns `IAP_FAILED` — same argument, same predicate, nothing touches `sector_size` in between. So `result = IAP_NACK` at 271 is dead. Harmless as defence in depth; I mention it only because it reads like a second independent check and isn't one.
2. **`iap_clear_upgrade_flag()` runs before the erase** (267 vs 268). Safe today precisely because of (1) — the erase cannot fail there. If it ever gains a real failure mode, that ordering leaves the flag cleared with nothing written. Probably worth a comment rather than a reorder.
3. **`test_bootloader_nacks_misaligned_erase_addresses` matches source strings** rather than exercising behaviour, so it passes on wrong logic and breaks on a rename. It's consistent with the test directly above it, so I'd leave it — but it pins the diff rather than the invariant.

All three are follow-up material. **From my side this is ready to merge** — the last open item is closed and everything I can check off-device is green.

Worth saying plainly: nineteen commits and +6,100 lines, and the review cycle on it has been you turning findings around in hours while I took thirteen days to look. The dual-boot switcher is the piece that unblocks adoption for everyone who hesitates at "can I get back to stock?", and it's in good shape. Thank you.

One flag before this goes anywhere near hardware — mine, not yours: I can't smoke-test the flash path on bench unit #1 right now, because the FPGA config-entry work needs the factory IAP MSC channel intact and reflashing the bootloader would take it out. So this merge is verified by build and test suite, not by a flash on my bench. If you've run `11e4f76` on your own device, say so on the thread and I'll note it in the merge.

Komzpa:
@DavidClawson I ran current head `52a61d1` on real 2C53T hardware.

Local build/test pass before hardware:

- `python3 scripts/test_flash_switcher.py`
- `python3 scripts/test_flash_preflight.py`
- `python3 scripts/test_bootloader_power_entry.py`
- `python3 scripts/test_bootloader_updater.py`
- `python3 scripts/test_stock_dispatcher_power_handoff.py`
- `python3 scripts/test_stock_hybrid_power_hold_patch.py`
- `python3 scripts/test_stock_settings.py`
- `python3 scripts/test_stock_settings_diff.py`
- `make -C firmware`
- `make -C firmware/bootloader`
- `make -C firmware/bootloader high`
- `make -C firmware/stage0`
- `make -C firmware/stock_dispatcher`
- `make -C firmware/bootloader_updater`
- `make -C firmware/high_layout_updater`

Hardware pass:

- entered high HID recovery as `2e3c:af01 Artery HID IAP`
- saved the stock settings page at `0x08006000`
- ran `python3 scripts/switch_firmware.py stock --flash`; it wrote 898 blocks and verified all CRCs:
  - `0x1A25DF03`
  - `0x2E979C7C`
  - `0x048D4C65`
- re-entered recovery and verified the settings page was an exact byte match to the saved baseline
- ran `python3 scripts/switch_firmware.py openscope`; it wrote 498 blocks and verified CRCs `0x30959EEF` and `0x67FB2634`
- OpenScope came up on CDC as `2e3c:5740 AT32 Virtual Com Port`; `status` reported `SYSCLK 240MHz`, FPGA initialized, and H2 upload done
- `reboot bootloader` from CDC returned the unit to HID recovery
- readback checks passed for the app, stock/settings page, and high recovery; the upgrade flag read back as `0x41544B38`, and settings still exactly matched the saved stock baseline

After the test I restored my unit back to stock with the same current head: `stock_user_dispatcher.bin` sha256 `6caafb136e623040a7f804fead9d9a653549fee5ef8a8eb939a62c351a3e3d77`, 898 blocks, the same three CRC matches above, `Flash verified`, then jump to `0x08000000`.

So yes: this has now been smoke-tested on real hardware both directions, stock settings survived the round trip, OpenScope booted far enough to enumerate and report status, and the device is back on stock at the end. I did not find a new blocker in the hardware pass.

DavidClawson:
@Komzpa — re-verified at `52a61d1` (not inheriting the `11e4f76` approval, since the head moved after it and the delta was not cosmetic). **Merging.**

Ubuntu 24.04, `arm-none-eabi-gcc` 13.2.1, isolated worktree.

## Builds — 7/7

| Target | Size |
|---|---|
| `firmware.bin` | 504,844 B — fits the new 932K `FLASH_APP` with **457,716 B spare** |
| `bootloader.bin` | 12,248 B — **12,188 / 16,384 used (74%)**, +92 B over `11e4f76` |
| `bootloader_high.bin` | 12,692 B |
| `stage0` / `stock_dispatcher` | 1,076 / 1,256 B |
| `bootloader_updater` / `high_layout_updater` | 26,008 / 27,832 B |

No new source warnings.

## Tests — 8/8 suites, **51 tests, 0 skips**

Worth noting for anyone reproducing: my first run reported 5 skips, all environmental (missing `archive/2C53T Firmware V1.2.0/` and unbuilt `.bin`s). Linking the stock archive and building first takes it to zero. One of those skips was `test_updater_binaries_gap_fill_stock_settings_page` — the only test in the suite that inspects a **real built binary** rather than source text, so it's the one you least want silently skipped. Separately, `test_stock_hybrid_power_hold_patch.py` raises a bare `FileNotFoundError` traceback when artifacts are missing where its siblings `skipTest` cleanly; a graceful skip there would save someone a confusing minute.

## The guard — verified by executing it, not reading it

I pulled `iap_erase_sector_base` / `iap_write_address_valid` / `iap_erase_required` **verbatim** out of `hid_iap_user.c`, compiled them on the host with the real constants (`SECTOR_SIZE_2K`, `HID_IAP_BUFFER_LEN = 1024`), and drove real address sequences through a 1:1 transcription of the `iap_address()` decision:

| Case | Result |
|---|---|
| Contiguous 498 × 1KB (your `openscope` run) | **0 NACKs, exactly 249 erases**, every byte lands on erased-and-not-yet-written flash ✅ |
| The **`11e4f76`** predicate | **would NACK 249 of 498 blocks** ✅ regression confirmed |
| Partial-preserve: `0x08006400` after `0x08005800` | NACK ✅ |
| Sub-1KB alignment (`0x08004200`) | NACK ✅ |

**Two things I want to say plainly about this.**

**First — you caught a bug in a change I approved, and my approval was the thing at fault, not your patch.** `11e4f76` gated *writes* on the *erase*-alignment predicate. On a 2K-sector part with `hid_flash.py` sending an ADDR per 1KB chunk, that is exactly half of all address packets. It would have failed on block 2 of every flash. It passed a clean build, 49 tests, and an explicit regression hunt on my side — because **every check I ran was a check against what the code says, not what it does.** The suite greps for source strings, so it will happily pin incorrect logic as correct. That's on my review process, and I'm sending a follow-up PR that turns the simulation above into an actual test.

**Second — this closes a caveat I filed on 28 July and then didn't follow through on.** I noted then that `iap_erase_sector()` was unchanged, so the "no partially-preserved sector" invariant lived only in the host tooling (`validate_preserved_sectors()`), not on the device. The partial-preserve row above is that invariant now enforced device-side. Nice.

## Two notes, neither blocking

**1. `switch_firmware.py` now requires `uv`, undocumented.** `HID_FLASH_CMD = ["uv", "run", str(HID_FLASH)]` replaced `sys.executable`. The script now uses **two different interpreters**: preflight still runs under `sys.executable`, the flash step runs under `uv run`. I confirmed the failure mode with `uv` off `PATH` — preflight passes and prints success, then it dies with `FileNotFoundError` at the moment it would write flash, which is the worst place to discover a missing dependency. `uv` isn't mentioned in `README.md`, `CONTRIBUTING.md`, `AGENTS.md` or `docs/`. Either document it as a requirement or fall back to `sys.executable` when `uv` isn't found — your call, and happy to take a patch either way.

**2. Latent (host-unreachable) hazard: a repeated ADDR to the same non-sector-aligned block.** `iap_erased_sector_address` stays set for the rest of the session, so a second ADDR to e.g. `0x08004400` is ACKed with **no erase** — the subsequent 1KB would be programmed over already-programmed flash, and since bits only go 1→0 that's silent corruption rather than an error. **Not reachable through `hid_flash.py`**: it has no retry path, it raises on NACK or timeout, and `verify_flash_crc` would catch the damage afterward. So this is a note for any future host that retries a block, not something to fix now. If you ever want it belt-and-braces, clearing `iap_erased_sector_address` once a sector's second half has been written would do it.

## Hardware

Your round trip at `52a61d1` is what closed this out — I approved on build+suite only and explicitly wouldn't flash bench unit #1, since its factory IAP MSC channel has to survive for the FPGA config-entry work. Stock settings coming back byte-exact across a full stock → OpenScope → stock cycle is the result that matters most for #9, and you restored your own unit to stock afterward without being asked.

I also verified the `stage0` flag fix in the **binaries** rather than the source: bootloader writes `0x41544b38`, `stage0` and `stock_dispatcher` both read `0x41544b38`. `0x41544F4B` was never going to match — stage0 could not have recognised a validated app.

Thank you for this one. It's a large, careful piece of work, it directly unblocks #9, and the turnaround on every round of review has been faster than mine.

Stlkv:
Not sure if this should be posted here, or if it’s useful at all. I asked Claude how I could help the project, and it suggested making some dumps, so here they are. [2c53t_factory_data_V14_unit2.zip](https://github.com/user-attachments/files/30975941/2c53t_factory_data_V14_unit2.zip) And here’s Claude’s note:

Following up on the open question from the review above ("any device that has ever booted stock should already have 224KB SRAM ... if anyone reading this has a stock unit and is willing to check, I'd like to close that loop"):

Unit: 2C53T, board rev V1.4, shipped with stock V1.2.0, case never opened, ROM DFU never touched. Method: flashed the `make guest` build via the factory IAP channel (MENU+Power, `iap_flash.py`), collected everything over the CDC debug shell, then restored stock V1.2.0 (sha256 `a17c5c35…` matches your reference).

**1. EOPB0 is already 0xFE on a stock-booted unit — no case opening required.** Two independent proofs:
- `FLASH_USD` register (`0x4002201C`) reads `0x03FFFFFE` — low byte `0xFE`, i.e. the loaded EOPB0 value is 0xFE and extended SRAM is active.
- Functional: the guest build carries ~217 KB of BSS; it boots and runs fine. On a 96 KB part it would HardFault before reaching the shell.

Caveat for anyone re-verifying: a **direct read of the USD region at `0x1FFFF800` returns all-FF** on this unit — FAP is enabled from the factory (consistent with your RDP-kills-SWD finding), and it blinds the direct readback. Use the `FLASH_USD` register or the functional test instead.

**2. The archived factory bootloader generalizes to a second V1.4 unit.** Dumped my unit's `0x08000000–0x08007000` via `dump_factory_bootloader.py`: code is **byte-identical** to `archive/factory_iap_bootloader_2C53T.bin` — the only differences are 355 bytes confined to `0x6002–0x6200`, which is the stock settings page at `0x08006000`, not code. The "this dump is from one V1.4 unit and may not match yours" disclaimer can be softened to "confirmed on two units".

**3. Full 16 MB W25Q128 dump included** — taken twice back-to-back, identical sha256 both times; ~80% blank. Offered for cross-unit diffing against bench unit #1 — e.g. for the factory-calibration hunt.

Everything is in the attached archive (both raw `.bin` dumps, the shell output, and a README with unit details and sha256 checksums).

Happy to run more checks on this unit if useful.

DavidClawson:
@Stlkv the unit-2 factory dumps are exactly the loop-closer I asked for in the review — EOPB0=0xFE confirmed on a never-opened stock unit (and the FLASH_USD-register method noted, since FAP blinds the direct read), factory bootloader byte-identical on a second V1.4 board. I'll soften the archive disclaimer to 'confirmed on two units' and the 16MB flash dump is queued for cross-unit cal diffing. Thank you — and given what you went on to do in #18 the same day, quite the debut.

PR #24, opened 2026-08-15
[WIP] Reconstruct editable GW1N-2 FPGA RTL

Komzpa:
> [!WARNING]
> This is an intermediate reconstruction draft. It is **not ready to merge** and has not been validated on hardware.

## What this draft contains

- Editable, modular SystemVerilog for the GW1N-2/GW1N-UV2 FPGA: capture buffers, runtime SPI, trigger/timebase, and an integration top.
- Grounded `0x04`/`0x05` runtime reads with observed 1026-byte CS-low framing; unknown commands and status values stay raw.
- Low-five-bit opcode decode matching the exhaustive alias sweep, a control-register bank committing the observed two-byte write frames, the bench-proven digital post-ADC trigger comparator on register `0x08` (stock arms `0xAD`), and an SPI-loadable rate divider reconstructing the traced `BSRAM_1/2` counter-gate shape with its load select marked as an explicit placeholder.
- A first buildable exact-part validation image (`debugclk_hw_top` + a six-pin QN48 constraint set + `build_debugclk_image.sh`): MCU-driven debug clock on IOB7B (pin 30), IOR1B run line (pin 46), and the four netlist-traced runtime SPI pins, with synthetic ramp / walking-one ADC data through the real capture path. SRAM-load-only, reproducible, no bitstream committed, unproven on hardware.
- Two inferred Gowin `DPX9B` capture memories at the default `1024x8` depth instead of register-expanded RAM.
- Simulations, a BSRAM synthesis regression, and structural-netlist comparison tools.
- Provenance for the stock payload at offset `0x4AD19`, length `115638`, and IDCODE `0x0120681B`.

This is reconstruction work, not recovered original HDL. It builds on [DavidClawson/OpenScope-2C53T#18](https://github.com/DavidClawson/OpenScope-2C53T/issues/18) and [YosysHQ/apicula#515](https://github.com/YosysHQ/apicula/issues/515).

## Checks run on `25eb00b`

- Python: 6 tests passed, including BSRAM/default-depth, changed-primitive, and wrong-input-hash negative controls.
- Icarus Verilog: all eight testbenches passed, including the comparator, control-register, and rate-divider units, the reworked integrated top, and the debug-clock hardware wrapper validated through its six physical nets only.
- Verilator 5.048: whole-RTL lint passed.
- Yosys 0.66 `synth_gowin`: two `DPX9B`, 84 `LUT4`; no memory-to-register warning.
- nextpnr-himbaechel (gw1n2 branch) on `GW1N-UV2QN48XFC6/I5` with the six-pin CST: placed on the exact constrained pads, routed, timing >90 MHz on both clock domains; `gowin_pack` emits a `.fs` carrying IDCODE `0x0120681B`, sha256-stable across independent builds (`9c465d72a5cd5866b7c4b06baa87f10b2e4e3c5f965f09c1518fa08c47d10751`).
- `git diff --check`: passed.

Exact-QN48 toolchain work lives at [Komzpa/apicula `agent/gw1n2-qn48-chipdb-20260815`](https://github.com/Komzpa/apicula/tree/agent/gw1n2-qn48-chipdb-20260815), now at commit `d978cad`. It builds a database containing `GW1N-UV2QN48XF -> QN48/GW1N-2` with the UV1P5 proxy absent, keys the package by its full ordering code so `gowin_pack` resolves it, and maps the top-right corner tile to `IOR1` (the run pin) instead of a nonexistent `IOT20`. With those fixes the debug-clock validation image builds end to end for the exact part.

## Why this remains a draft

- Exact board-net constraints are complete only for the six netlist-traced pins used by the debug-clock image; the ADC bus, PLL clock input, and remaining enable mappings still lack continuity-grade evidence, so the full scope top still cannot be placed.
- The stock rPLL configuration remains undecoded; the buildable image substitutes an MCU-driven debug clock and synthetic ADC data, so it validates the capture/trigger/SPI fabric, not stock sample rates.
- The archived Saleae `.sal` trace is hash-pinned, but Logic 2/export tooling is unavailable here, so byte-for-byte trace replay is not yet automated.
- The trigger threshold register (`0x08`, in ADC codes) is bench-decoded, but stock source/edge selection, hysteresis, holdoff, pre/post-trigger policy, buffer ordering, CDC behavior, the rate-divisor register address, and status-byte semantics remain unknown.
- No formal or behavioral equivalence to the stock design is claimed.
- No FPGA image from this branch has been loaded onto hardware.

The branch contains no generated bitstream and does not authorize non-volatile FPGA programming.

### DIFF
--- a/fpga/README.md
+++ b/fpga/README.md
@@ -1,0 +1,47 @@
+# FPGA reconstruction workspace
+
+This directory is the editable, evidence-tracked reconstruction of the FNIRSI
+2C53T scope FPGA. It is not the original vendor RTL.
+
+The target is the physically identified Gowin **GW1N-UV2 in QN48**, a member of
+the **GW1N-2** family. The stock configuration stream carries IDCODE
+`0x0120681B`. Package identity and device identity are established; board signal
+pin assignments are deliberately not copied here until the official QN48 table
+and board continuity results are integrated together.
+
+## Layout
+
+- `rtl/` — small, reviewed SystemVerilog blocks reconstructed from observed
+  behavior and structural evidence.
+- `sim/` — focused simulation testbenches for those blocks.
+- `tools/` and `tests/` — structural-comparison helpers and their tests.
+- `docs/provenance.md` — immutable stock inputs, hashes, and tool revisions.
+- `docs/reconstruction.md` — recovered block map, protocol facts, and unknowns.
+- `docs/verification.md` — proof levels and acceptance criteria.
+
+`rtl/capture_channel.sv` is currently a portable circular capture-memory
+primitive. Its unit test proves wrap, freeze, readback, and the disabled-capture
+negative control. It does **not** yet prove Gowin BSRAM inference, stock memory
+ordering, trigger behavior, timing closure, or hardware equivalence.
+
+## Quick local checks
+
+When Icarus Verilog and Verilator are installed:
+
+```sh
+iverilog -g2012 -s tb_capture_channel \
+  -o /tmp/tb_capture_channel \
+  fpga/rtl/capture_channel.sv fpga/sim/tb_capture_channel.sv
+vvp /tmp/tb_capture_channel
+verilator --lint-only --timing -Wall fpga/rtl/capture_channel.sv
+python3 -m unittest fpga/tests/test_netlist_verification.py
+```
+
+These are simulation and helper checks only. See
+[`docs/verification.md`](docs/verification.md) before making a stronger claim.
+
+## Safety
+
+All target experiments must load volatile **SRAM only**. Never write the FPGA's
+internal non-volatile flash: it contains the stock meter design and no verified
+recovery image exists. No file in this directory authorizes flashing hardware.
--- a/fpga/README.md
+++ b/fpga/README.md
@@ -29,8 +29,11 @@ the sample-clock/SPI-clock CDC protocol.
 
 The unit test proves wrap, freeze, readback, and the disabled-capture negative
 control. The BSRAM check proves the target synthesis netlist retains two `DPX9B`
-cells and rejects a register fallback. Neither check proves stock memory order,
-trigger behavior, timing closure, or hardware equivalence.
+cells and rejects a register fallback. The register, comparator, and divider
+blocks carry their own testbenches for the bench-grounded runtime writes,
+digital trigger level, opcode aliasing, and divided-tick contracts. None of
+these checks proves stock memory order, trigger behavior, timing closure, or
+hardware equivalence.
 
 ## Quick local checks
 
@@ -43,7 +46,9 @@ iverilog -g2012 -s tb_capture_channel \
 vvp /tmp/tb_capture_channel
 verilator --lint-only --timing -Wall \
   fpga/rtl/capture_channel.sv fpga/rtl/trigger_timebase.sv \
-  fpga/rtl/spi_runtime_interface.sv fpga/rtl/fnirsi_2c53t_top.sv
+  fpga/rtl/trigger_comparator.sv fpga/rtl/spi_runtime_interface.sv \
+  fpga/rtl/spi_control_registers.sv fpga/rtl/rate_divider.sv \
+  fpga/rtl/fnirsi_2c53t_top.sv
 python3 -m unittest fpga/tests/test_capture_bsram_mapping.py
 python3 -m unittest fpga/tests/test_netlist_verification.py
 ```
--- a/fpga/README.md
+++ b/fpga/README.md
@@ -35,6 +35,18 @@ digital trigger level, opcode aliasing, and divided-tick contracts. None of
 these checks proves stock memory order, trigger behavior, timing closure, or
 hardware equivalence.
 
+## Debug-clock hardware image
+
+`rtl/debugclk_hw_top.sv` plus `constraints/fnirsi_2c53t_qn48.cst` form the
+first buildable image for the exact GW1N-UV2/QN48 part. It uses only the six
+pins with netlist-grade evidence (runtime SPI, the IOR1B run line, and the
+clock-capable IOB7B pad as an MCU-driven debug sample clock) and feeds the
+capture path synthetic on-chip ramp and walking-one data, so no unproven ADC,
+clock, or board net is touched. `tools/build_debugclk_image.sh` reproduces the
+`.fs` build (synthesis, exact-part place-and-route, packing) with pinned tools;
+build products stay out of the tree. The image is for volatile SRAM loading
+only and has not yet been proven on hardware.
+
 ## Quick local checks
 
 When Icarus Verilog and Verilator are installed:
@@ -48,7 +60,7 @@ verilator --lint-only --timing -Wall \
   fpga/rtl/capture_channel.sv fpga/rtl/trigger_timebase.sv \
   fpga/rtl/trigger_comparator.sv fpga/rtl/spi_runtime_interface.sv \
   fpga/rtl/spi_control_registers.sv fpga/rtl/rate_divider.sv \
-  fpga/rtl/fnirsi_2c53t_top.sv
+  fpga/rtl/fnirsi_2c53t_top.sv fpga/rtl/debugclk_hw_top.sv
 python3 -m unittest fpga/tests/test_capture_bsram_mapping.py
 python3 -m unittest fpga/tests/test_netlist_verification.py
 ```
--- a/fpga/README.md
+++ b/fpga/README.md
@@ -19,10 +19,18 @@ and board continuity results are integrated together.
 - `docs/reconstruction.md` — recovered block map, protocol facts, and unknowns.
 - `docs/verification.md` — proof levels and acceptance criteria.
 
-`rtl/capture_channel.sv` is currently a portable circular capture-memory
-primitive. Its unit test proves wrap, freeze, readback, and the disabled-capture
-negative control. It does **not** yet prove Gowin BSRAM inference, stock memory
-ordering, trigger behavior, timing closure, or hardware equivalence.
+The top-level default has two 1024 x 8 raw capture stores. The portable
+registered-read/write split maps through the installed Yosys Gowin mapper to
+two `DPX9B` BSRAM primitives, one per channel. Each store needs 8192 bits; the
+mapper's 14-bit, byte-addressed BSRAM primitive has 18432 initialization bits,
+so the two stores need two primitives rather than one shared store. This is a
+resource-mapping result, not a claim about stock order, trigger behavior, or
+the sample-clock/SPI-clock CDC protocol.
+
+The unit test proves wrap, freeze, readback, and the disabled-capture negative
+control. The BSRAM check proves the target synthesis netlist retains two `DPX9B`
+cells and rejects a register fallback. Neither check proves stock memory order,
+trigger behavior, timing closure, or hardware equivalence.
 
 ## Quick local checks
 
@@ -33,7 +41,10 @@ iverilog -g2012 -s tb_capture_channel \
   -o /tmp/tb_capture_channel \
   fpga/rtl/capture_channel.sv fpga/sim/tb_capture_channel.sv
 vvp /tmp/tb_capture_channel
-verilator --lint-only --timing -Wall fpga/rtl/capture_channel.sv
+verilator --lint-only --timing -Wall \
+  fpga/rtl/capture_channel.sv fpga/rtl/trigger_timebase.sv \
+  fpga/rtl/spi_runtime_interface.sv fpga/rtl/fnirsi_2c53t_top.sv
+python3 -m unittest fpga/tests/test_capture_bsram_mapping.py
 python3 -m unittest fpga/tests/test_netlist_verification.py
 ```
 
--- a/fpga/constraints/fnirsi_2c53t_qn48.cst
+++ b/fpga/constraints/fnirsi_2c53t_qn48.cst
@@ -1,0 +1,14 @@
+// GW1N-UV2 QN48 physical constraints for debugclk_hw_top.
+//
+// Pin numbers are the audited UG171 QN48 package table as built into the
+// apicula agent/gw1n2-qn48-chipdb-20260815 database (GW1N-UV2QN48XF), NOT the
+// GW1N-1P5C QFN48XF proxy numbers.  IOB locations are the netlist-traced
+// runtime-SPI and run nets from the stock design.  Only these six nets have
+// continuity-grade or netlist-grade evidence; no other board pin may be added
+// here without the same standard of proof.
+IO_LOC "dbg_clk" 30;
+IO_LOC "run_enable" 46;
+IO_LOC "spi_cs_n" 34;
+IO_LOC "spi_sclk" 29;
+IO_LOC "spi_mosi" 35;
+IO_LOC "spi_so" 28;
--- a/fpga/docs/provenance.md
+++ b/fpga/docs/provenance.md
@@ -53,6 +53,34 @@ Pin every reproduction rather than following moving branch heads:
 | GW1N-2 place-and-route development | `DavidClawson/nextpnr`, `gw1n2` | `59c8f93a62ee17adc8d1c5d7f399697ffeb85c0f` |
 | exact GW1N-UV2/QN48 chipdb, packing, and the debug-clock image build | `Komzpa/apicula`, `agent/gw1n2-qn48-chipdb-20260815` | `d978cad` (atop `cea1618`) |
 
+## Debug-clock image reproduction
+
+Commit `25eb00b` is the first buildable exact-part debug-clock image. The
+recorded build used Yosys `0.66` (`git sha1 86f2ddebce7e98ce7cacc27e8a5c14cb53b51b51`) and Python `3.14.7`.
+
+| Artifact | SHA-256 |
+|---|---|
+| `apycula/GW1N-2.msgpack.xz` built at `Komzpa/apicula` `d978cad` | `3dc7b45d3c2eb1dc714a0e9bbbb55c2ab154d6a5bc6d546d8c947c8e452e285e` |
+| `debugclk_hw_top.fs` from the recorded build | `9c465d72a5cd5866b7c4b06baa87f10b2e4e3c5f965f09c1518fa08c47d10751` |
+
+Reproduce it with (all four tools from the pinned revisions in the table
+above):
+
+```sh
+APICULA_PYTHON=<python with apycula installed -e at d978cad> \
+NEXTPNR=<nextpnr-himbaechel built from DavidClawson/nextpnr gw1n2 59c8f93a> \
+NEXTPNR_SRC=<that nextpnr source tree> \
+BBASM=<bba/bbasm from the same nextpnr build> \
+OUT=<empty build dir> \
+sh fpga/tools/build_debugclk_image.sh
+sha256sum "$OUT"/debugclk_hw_top.fs
+```
+
+At the pinned baseline the command used the exact-part Apicula checkout
+providing `apycula/GW1N-2.msgpack.xz`; without that chipdb,
+`nextpnr-himbaechel` reports `No package for partnumber GW1N-UV2QN48XFC6/I5`.
+Keep the hash tied to this commit: later RTL changes change the `.fs` image.
+
 The chipdb build requires Gowin EDA device data for `GW1N-1P5C`:
 `GW1N-1P5C.fse`, `.dat`, and `.tm`. Those vendor inputs are not checked in.
 Without the exact generated `GW1N-2.msgpack.xz`, a fresh unpack is blocked and
--- a/fpga/docs/provenance.md
+++ b/fpga/docs/provenance.md
@@ -51,6 +51,7 @@ Pin every reproduction rather than following moving branch heads:
 | stock unpack and GW1N-2 chipdb baseline | `DavidClawson/apicula`, `gw1n2-build` | `2987a6a323107e54d02b2d47a452f0fd09c95a07` |
 | GW1N-2 pinout/IO development | `DavidClawson/apicula`, `gw1n2-io` | `cf5ab40d203a3f79dc233bdd8000b7c5fc2dfacb` |
 | GW1N-2 place-and-route development | `DavidClawson/nextpnr`, `gw1n2` | `59c8f93a62ee17adc8d1c5d7f399697ffeb85c0f` |
+| exact GW1N-UV2/QN48 chipdb, packing, and the debug-clock image build | `Komzpa/apicula`, `agent/gw1n2-qn48-chipdb-20260815` | `d978cad` (atop `cea1618`) |
 
 The chipdb build requires Gowin EDA device data for `GW1N-1P5C`:
 `GW1N-1P5C.fse`, `.dat`, and `.tm`. Those vendor inputs are not checked in.
--- a/fpga/docs/provenance.md
+++ b/fpga/docs/provenance.md
@@ -1,0 +1,71 @@
+# Provenance and reproducibility
+
+## Immutable stock inputs
+
+The canonical source is the stock MCU application archive, not a generated
+Verilog file:
+
+| Artifact | Size | SHA-256 | Provenance |
+|---|---:|---|---|
+| `APP_2C53T_V1.2.0_251015.bin` | 751232 bytes | `a17c5c35c97bb898f15672a1747bc1041d8ed507c16999ddba0d1e4e2ec0c760` | FNIRSI V1.2.0 application archive |
+| raw FPGA slice | 115638 bytes | `5a0e73384e496bdb3b3d591b852bec2e806e70cbc71439c9829695324efd5c3b` | exact bytes at file offset `0x4AD19` in the archive |
+
+Reproduce the second hash without creating a file:
+
+```sh
+dd if='APP_2C53T_V1.2.0_251015.bin' bs=1 \
+  skip=$((0x4AD19)) count=115638 status=none | sha256sum
+```
+
+The stream header identifies the device family with IDCODE `0x0120681B`.
+Physical inspection identifies the installed part as `GW1N-UV2`, package
+`QN48`; `GW1N-UV2` uses the GW1N-2 fabric/tool device.
+
+The stock payload and generated netlist may retain FNIRSI rights. Their presence
+or reproducibility does not grant a redistribution license. Keep provenance and
+licensing separate.
+
+## Published reference artifacts
+
+The immutable published reference is
+`DavidClawson/gw1n2-apicula@7e44f773ccdabe8cce16b13d85b491e1321c8949`:
+
+| Path at that commit | Size | SHA-256 | Meaning |
+|---|---:|---|---|
+| `tools/m5/scope.fs` | 925842 bytes | `f2fde7b6458bacb4d8d607e97277531e36823a715bcb2bee11380a082742fd44` | ASCII-form configuration stream; decoded payload matches the raw FPGA-slice hash above |
+| `tools/m5/scope_unpacked.v` | 2179558 bytes | `982a94ed69cadb16b45a18a06bc3412895a933f6e753f3260a16a1c8aca9ee32` | Apicula-generated primitive netlist |
+
+`scope_unpacked.v` is staged evidence, not source truth. It is a single anonymous
+top-level fabric netlist made of Gowin primitives. It preserves useful
+connectivity, LUT initialization, memories, and I/O-cell structure, but not the
+original hierarchy, names, comments, intent, clock constraints, or HDL. Some
+primitive inputs and PLL semantics are not recovered. Editing it would create a
+gate patch, not maintainable reconstructed RTL.
+
+## Tool revisions
+
+Pin every reproduction rather than following moving branch heads:
+
+| Purpose | Repository branch | Commit |
+|---|---|---|
+| stock unpack and GW1N-2 chipdb baseline | `DavidClawson/apicula`, `gw1n2-build` | `2987a6a323107e54d02b2d47a452f0fd09c95a07` |
+| GW1N-2 pinout/IO development | `DavidClawson/apicula`, `gw1n2-io` | `cf5ab40d203a3f79dc233bdd8000b7c5fc2dfacb` |
+| GW1N-2 place-and-route development | `DavidClawson/nextpnr`, `gw1n2` | `59c8f93a62ee17adc8d1c5d7f399697ffeb85c0f` |
+
+The chipdb build requires Gowin EDA device data for `GW1N-1P5C`:
+`GW1N-1P5C.fse`, `.dat`, and `.tm`. Those vendor inputs are not checked in.
+Without the exact generated `GW1N-2.msgpack.xz`, a fresh unpack is blocked and
+must not be reported as reproduced.
+
+The expected unpack command at the pinned baseline is:
+
+```sh
+PYTHONPATH=/path/to/apicula python3 -m apycula.gowin_unpack \
+  -d GW1N-2 scope.fs -o scope_unpacked.v
+sha256sum scope_unpacked.v
+```
+
+An exact byte hash is the strongest reproduction result. If emission order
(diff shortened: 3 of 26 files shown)


DavidClawson:
This is a serious piece of work — thanks for opening it as a draft with the caveats spelled out; that's exactly the right framing.

Two pointers so the efforts converge rather than diverge:

- Most of the "unknown" semantics you list are already partially decoded in **[gw1n2-apicula](https://github.com/DavidClawson/gw1n2-apicula)** (sibling repo). Specifically: the capture arm/run register (an SI-data flop in the R16C8/9 bank, gated by IOR1B=PB11 ∧ IOB7B=PC6 — the netlist trace behind `guest-coldtrace` arming), the two DPX9B capture memories with their geometry and ADC bit order, and `tools/m_capture.py`, which captures synthetic ADC data end-to-end in sim from the *actual* payload. Also `m_simarm.py` (gate-level arm sim). If your `capture_channel.sv` / `spi_runtime_interface.sv` were checked against those rather than against the `.sal` framing alone, we'd know quickly where the reconstruction and the real fabric agree.
- Bench facts from this week that constrain the trigger/timebase model: the trigger comparator is **digital, post-ADC** — SPI register `0x08` is the trigger level in ADC codes (stock arms `0xAD`); the default capture rate measured today is **~1.07 kS/s** (1024-sample buffer ≈ 0.96 s), and stock runs ~23× faster by a rate control we believe lives on BSRAM_1/2 and is SPI-reachable but haven't found the opcode for yet. A reconstruction that produces a programmable-rate counter there would be a real lead.

I'll leave this as draft and read it properly rather than skim-merging — the structural verification tools alone look worth having in-tree. Fully agree on no NV programming; nothing here should touch the FPGA flash until we can read it back.

Komzpa:
Thanks — both pointers acted on, pushed as `26001fe`.

**Bench facts folded in:**

- The trigger comparator is now in the fabric as `trigger_comparator.sv`: digital, post-ADC, comparing codes against SPI register `0x08`, reset default `0xAD`. Rising-crossing is the declared local contract; edge/source/hysteresis/holdoff stay listed as unknowns since the noise-band experiment can't separate polarities.
- The runtime write path exists now: `spi_control_registers.sv` commits the observed two-byte CS frames at the CS boundary, with the five arm-sequence values (`01 08 / 02 03 / 06 00 / 07 00 / 08 AD`) as reset defaults. `0x01/0x02/0x06/0x07` are stored and re-exposed raw.
- Fixed a real divergence from your sweep: the SPI decoder compared full opcode bytes, contradicting the exact `0x20–0x3F` → `0x00–0x1F` aliasing. It now decodes only the low five bits, and `0x24`-serves-CH1 is a testbench regression in both the unit and top-level sims.
- The programmable-rate counter lead: `rate_divider.sv` reconstructs the M12 shape — a counter with an SPI-side load path gating the slow-path write cadence. The load select is an explicit placeholder (`0x0B`): our byte-exact arm replay stayed at the slow default (`trigger_regime_findings_2026-08-14.md`, Addendum 2 in this branch), which argues the five known registers are not the divisor, so I deliberately picked an unobserved address and marked it as such. The top exports the divided tick instead of instantiating a `BSRAM_1/2` store pair, because that record's function (accumulate / min-max / decimation) is still only the M9 hypothesis.

**On cross-checking against `m_capture.py` / `m_simarm.py`:** the public [gw1n2-apicula](https://github.com/DavidClawson/gw1n2-apicula) repo's last push is the June 14 snapshot — those two tools and the capture-memory geometry / ADC-bit-order decode you mention aren't in it. The RTL was checked against what is public (the R3 trace: always-write-enabled BSRAMs, counter-window CEA, the SPI-register bit in the arm path, four-way SO mux). If you can push the newer tools, the next step on our side is wiring the structural-netlist comparison against the actual-payload sim instead of the `.sal` framing alone.

**One measurement to reconcile:** your ~1.07 kS/s today vs the ~2.4–2.7 kS/s we got on 2026-08-14 by two methods (trigger-frequency cutoff at ~2.8 Hz, and 30 Hz → ~13 cycles per 1024-sample buffer). More than a 2× spread between benches on the same default regime smells like a condition difference — range relay state or DAC1 posture, maybe. Both numbers are recorded in `fpga/docs/reconstruction.md` without picking a winner; the RTL encodes no clock frequency.

Checks re-run on `26001fe`: seven Icarus testbenches pass (three new blocks have their own), Verilator 5.048 whole-RTL `-Wall` lint clean, `synth_gowin` still maps exactly two `DPX9B` with no register fallback, both python suites pass, `git diff --check` clean. Still draft, still nothing here authorizes NV programming.

Komzpa:
Milestone: the branch now builds a loadable image for the exact part, end to end. Pushed as `25eb00b` here plus `d978cad` on [Komzpa/apicula `agent/gw1n2-qn48-chipdb-20260815`](https://github.com/Komzpa/apicula/tree/agent/gw1n2-qn48-chipdb-20260815).

**What it is.** `debugclk_hw_top` + `fpga/constraints/fnirsi_2c53t_qn48.cst` + `fpga/tools/build_debugclk_image.sh`: the reconstruction top wrapped for hardware using only the six pins with netlist-grade evidence — the runtime SPI four (QN48 pins 28/29/34/35), the IOR1B run line (pin 46), and IOB7B (pin 30, `GCLKC_4`) as an **MCU-driven debug sample clock**, sidestepping the undecoded rPLL. The ADC pins stay untouched: CH1 samples a synthetic on-chip ramp and CH2 a walking-one marker, through the real capture/trigger path. Yosys → nextpnr on `GW1N-UV2QN48XFC6/I5` (placed on exactly the constrained pads, routed, >90 MHz on both domains against MHz-at-most debug clocks) → `gowin_pack` → `.fs` carrying IDCODE `0x0120681B`. sha256 `9c465d72a5cd5866b7c4b06baa87f10b2e4e3c5f965f09c1518fa08c47d10751`, bit-identical across independent rebuilds; the script reproduces it, no bitstream is committed.

**Expected bench behavior** (all through the existing bit-bang SRAM path; PB11/PC6 are the same pins coldtrace already drives as arm co-signals):

- MCU toggles PC6 as the sample clock; the first eight edges release the internal power-on reset.
- PB11 high = run. The CH1 ramp crosses the default `0xAD` trigger level and freezes both buffers.
- `0x04` returns status `52 31 56` ("R1V", deliberately non-stock markers) + a frozen incrementing ramp; `0x05` the same markers + walking-one bytes; repeated reads byte-identical; unwritten words read `0x00` from configured BSRAM.
- PB11 low→high re-arms and refreshes. Writing `08 <level>` moves the freeze point — the register/comparator path live on hardware.
- This doubles as the R1 byte-mapping validator: known content through the real readout mux answers start-offset/stride/lane-order in one read.

SRAM only, as always: power cycle restores stock, and nothing in the branch authorizes NV programming.

**Two apicula fixes were needed** (both on the branch): the UV2 package is now keyed by its full ordering code with the speed grade, matching the other devices, so `gowin_pack` resolves the partnumber after a successful route; and the GW1N-2 top-right corner tile now maps to `IOR1` for bank lookups — it previously resolved to a nonexistent `IOT20B` and crashed packing, which is notable because that corner pad is precisely the scope's run input.

**Bonus finding for the DavidClawson/OpenScope-2C53T#18 JTAG hunt:** the audited UG171 QN48 column places TMS/TCK/TDI/TDO at pins **8/9/10/11**. The R3-era caveat ("pins 8–11 look like ordinary left-edge GPIO") was an artifact of the QFN48XF proxy bondout; on the real package the gold-pad assumption holds. That is direct support for the FT232H JTAG oracle route — JTAG SRAM load would bypass the config-entry question entirely.

DavidClawson:
Pushed — `DavidClawson/gw1n2-apicula@721b8d2` catches the public repo up through 2026-08-14. The two you named are there (`tools/m_capture.py`, `tools/m_simarm.py`) along with `m_divider.py` (the M12 pure-python micro-sim of the GB40 clock-gate divider — the closest thing to a spec for `rate_divider.sv`), `m_timebase.py`, `m_regime.py`, `m_readmux.py`, `m_spidomain.py`, `m_clocktree.py`, `m_countchk.py`, `m_gate40.py`, `m_segcheck.py`, `m_utilization.py`, plus the progress log. Sorry it was stale — that repo gets pushed in bursts.

**Set your `rate_divider.sv` load select to `0x01`, not `0x0B`.** Today's bench sweep found the timebase: the **low nibble of SPI register `0x01`** is a 1-2-5 sample-rate ladder, measured in triggered mode against a DDS reference —

| reg 01 | rate | | reg 01 | rate |
|---|---|---|---|---|
| `0F` | ≈30 kS/s | | `0B` | ≈500–600 kS/s |
| `0E` | ≈60 kS/s | | `0A` | ≈1 MS/s |
| `0D` | ≈150 kS/s | | bit 4 (`10`,`1F`) | ≈15 kS/s |
| `0C` | ≈300 kS/s | | | |

(≥300 kS/s are indicative — the reference is a 40 kSa/s software DDS up there.) So one of the five known arm registers **is** the divisor after all; the reason a byte-exact replay stayed slow is that the replay includes stock's `01 08`, and `08` is the slow default. This also retires the "23× faster" mystery: stock is almost certainly running `0F`.

**On the 1.07 vs 2.4–2.7 kS/s reconciliation — I think both are artifacts, and neither is a sample rate.** At reg01=`08` the zero-crossing count per buffer is **tone-independent**: 200 Hz and 1 kHz inputs both produced ~190 crossings per 1024-sample buffer, and triggered reads at `08` come back torn. A regime whose apparent period doesn't scale with input frequency isn't sampling in the ordinary sense — decimation with min/max or peak-detect would look exactly like this, which is interesting given where M9 put that record. On the `0F`…`0A` ladder the numbers behave properly (period scales with tone, frames are whole). So I'd treat `08` as a separate mode to characterise, not a rate to reconcile — and I'd build on `0F`.

**The `debugclk_hw_top` image is the instrument we need, and I'd like to run it.** Known content through the real readout mux answers the question that is blocking our display work: our reads come back with a **seam** — the engine free-runs with the run line held, every read starts at buffer address 0, and three consecutive frames share only ~20% of samples, so the oldest→newest boundary lands at a wandering index. Your `52 31 56` + ramp / walking-one would settle start-offset, stride and lane order in one read, and the PB11 low→high re-arm behaviour you predict is exactly the one-shot question. Two things before I flash it: (1) it goes in over the same bit-bang SRAM path, SRAM-only, power cycle restores stock — confirmed, that's the loader we use; (2) can you point me at the exact `build_debugclk_image.sh` invocation and the expected `.fs` sha256 so I can verify what I'm loading matches your `9c465d72…`?

Also noting your header-byte prediction against ours: we see the third stored MISO byte vary 0x50–0xA9 frame to frame while stock treats CH1 byte 2 == `0x01` as buffer-valid. If your image returns fixed markers there, that alone tells us whether those bytes are status or a write pointer.

**The QN48 8/9/10/11 JTAG correction is a big deal for #18** — thank you for auditing it rather than inheriting the proxy bondout. That makes the FT232H oracle a live option again: a JTAG SRAM load bypasses config entry entirely, and it's the one route we've never had a clean pin map for.

Komzpa:
`721b8d2` fetched — thank you, `m_divider.py` plus the bench ladder closes our biggest open question. The 0x0B placeholder is retired as of today: `rate_divider.sv` is being moved onto reg `0x01` with the 1-2-5 decode, and the docs lose the 1.07-vs-2.4 "conflict" in favor of your reading (both artifacts of mode `08`, which we'll carry as a distinct unproven mode rather than a ladder point). You were right that the replay argument was backwards — the replay *includes* `01 08`, so it proved nothing about reg 01. I'll push the rework to this branch once it passes the sim/synth gates.

**Build verification for `debugclk_hw_top`** — pin your build to `25eb00b` (the current branch head); later pushes will change the image. Invocation from the repo root:

```sh
APICULA_PYTHON=<venv-python>  NEXTPNR=<nextpnr-himbaechel>  \
NEXTPNR_SRC=<nextpnr-tree>    BBASM=<bbasm>                 \
  sh fpga/tools/build_debugclk_image.sh
```

with the inputs pinned as:

| Input | Pin |
|---|---|
| yosys | 0.66 (`86f2ddebce`) |
| apycula (venv, `pip install -e`) | `Komzpa/apicula` @ `d978cad` (`agent/gw1n2-qn48-chipdb-20260815`), Python 3.14.7 |
| chipdb `apycula/GW1N-2.msgpack.xz` | sha256 `3dc7b45d3c2eb1dc714a0e9bbbb55c2ab154d6a5bc6d546d8c947c8e452e285e` |
| nextpnr + bbasm | your `DavidClawson/nextpnr` `gw1n2` @ `59c8f93a` |

Expected: `sha256(debugclk_hw_top.fs)` = `9c465d72a5cd5866b7c4b06baa87f10b2e4e3c5f965f09c1518fa08c47d10751`. Re-confirmed today from a clean build directory — bit-identical. If your sha differs, compare the chipdb msgpack hash first: it's the only build input derived from vendor device files (`GW1N-1P5C.fse/.dat/.tm`, per `fpga/docs/provenance.md`), everything downstream is deterministic. Loader contract confirmed on our side too: SRAM-only over the bit-bang path, power cycle restores stock, nothing touches NV flash.

**What the image will answer, concretely.** Frame layout is: byte 0/1/2 = fixed markers `0x52 0x31 0x56` ("R1V"), samples from byte 3 onward, and every read starts at buffer address 0 by construction. CH1 is a free-running 8-bit ramp, CH2 is a walking one (`1 << ramp[2:0]`). So on your header-byte question: if byte 2 reads back as a constant `0x56`, that mux position is register-driven status — and stock's varying `0x50–0xA9` there is stock exposing something live (a write pointer being the obvious candidate). If it varies even on this image, that byte never passes through the response mux we reconstructed, and the model is wrong in a way worth knowing. The ramp gives you start-offset and stride in one read; the walking-one lane pins CH1/CH2 interleave. Re-arm is the one-shot experiment you described: run line (pin 46 / PB11) low→high re-arms, capture free-runs while it's high, sampling is paced entirely by your debug clock on pin 30 (PC6) — one sample per rising edge, so you control the effective rate.

One caveat to keep expectations honest: this image samples synthetic data through the real capture/readout path — it validates the mux, addressing, and re-arm story, but it deliberately claims nothing about the ADC pins, the rPLL, or stock status semantics.

We're also running your published `m_divider.py` against the netlist to try to reproduce the ladder structurally (drive the cluster externals with the reg-01 nibbles, read the GB40 gate period). If the divisor integers fall out cleanly I'll post them here — that would turn the measured ladder into an exact spec for `rate_divider.sv`.

Komzpa:
The reg `0x01` rework just landed on this branch as `61c88af`:

- `rate_divider.sv` decodes the low-nibble ladder into a divisor scaled off a parameterized `BASE_DIVISOR` — `0x0F` carries the largest ladder divisor, faster steps divide it (/2, /5, /10, /20, /33), `0x10`/`0x1F` double it. Only your measured ratios are claimed; the absolute base clock stays a parameter until the PLL is decoded.
- `0x08` is deliberately *not* a ladder point: it keeps the old slow-default divider and is documented as a distinct unproven mode (your tone-independent zero-crossing observation is recorded as the reason).
- The out-of-map placeholder select `0x0b` is gone; `spi_control_registers.sv` sources the rate select from stored reg `0x01`.
- `fpga/docs/reconstruction.md` retires the 1.07-vs-2.4–2.7 kS/s "conflict" (both artifacts of mode `08`) and the 23× mystery, and carries your ladder as a measured-rate table; `fpga/docs/provenance.md` now has the debug-clock reproduction recipe with the pinned hashes.

Gates on `61c88af`: all eight `fpga/sim` benches pass under iverilog, `synth_gowin` still infers exactly two DPX9B, and the image rebuilds — sha256 at this commit is `41ab45ddb131de4b361917fbaed59e6562344cec00876a5f7ffad3c5b0a6d7f5`. For your flash, `25eb00b` → `9c465d72…d10751` remains the pinned pair from my previous comment; both builds were byte-stable across repeated runs (comment-only RTL edits don't move the bitstream — the `.fs` came out identical before and after).

Closing the loop on the `m_divider.py` structural run: no divisor integers, and the reason matches what your progress log already records. The published stock runs give a flat GB40 (`ones=0/4000` under both all-zero and all-one external drives); the one SPI-loaded control bit we can name (`R16C9_DFFE_3`, per `m_rearm.py`/R3) sits *inside* the extracted 285-flop cluster, so `run(ext)` can't force it; and the `*_LB21` clears / `*_GB00` clocks are undriven long-wire branches in `scope_unpacked.v`. So the micro-sim confirms the gate cone reaches the SPI/arm bank but can't turn your bench ladder into integer divisors — your bench table plus the debugclk image on hardware stay the oracles, and that's how the RTL now states its confidence.

DavidClawson:
Rework noted, and thank you for stating the `m_divider.py` null result plainly rather than reaching for integers that weren't there — "your bench table plus the debugclk image on hardware stay the oracles" is the right confidence level for that RTL.

**I can answer your header-byte question directly, and the answer changes it.** I spent tonight decoding stock's runtime SPI3 dispatch out of the V1.2.0 image rather than inferring it from captures. The `fpga` task's loop is a 9-entry `tbh` jump table at `0x0803E536`, with `r7` = `0x40003C08` (SPI3 STS/DT), `r4` = `0x20002D78` (the `spi3_data` queue), `r9` = `0x200000F8` (scope state):

| op | handler | payload / behaviour |
|---|---|---|
| `0x01` | `0x0803E54C` | TX `state[0x2d]` = timebase index |
| `0x02` | `0x0803E970` | TX `state[0x14]` ∈ {1,2,3} = CH1 / CH2 / BOTH (else `3`) |
| `0x03` | `0x0803E9F2` | TX `0xFF`; shifts the roll-mode buffers |
| `0x04` | `0x0803E5A4` | TX `0xFF`, then 1024 × `0xFF` → `state[0x5B0 + 0..1023]` |
| `0x05` | `0x0803E68C` | TX `0xFF`, then 1024 × `0xFF` → `state[0x5B0 + 1024..2047]` |
| `0x06` | `0x0803ED1C` | TX `state[0x16]` (used elsewhere as an array index) |
| `0x07` | `0x0803ED5C` | TX `state[0x18]` ∈ {0..3} |
| `0x08` | `0x0803E75C` | TX the VFP-computed trigger comparator level |
| `0x09` | `0x0803E79C` | TX `FF FF`, keep byte[2]; then **CS toggle and a new frame with opcode `0x0A`**, keep byte[2] → a 16-bit value in `state[0x46]` |

**Stock discards exactly TWO bytes — the opcode echo and one dummy — then captures 1024 samples.** So there is no third header byte in stock's readout: what the June capture shows varying over `0x50–0xA9` at byte 2 is **sample[0]**, not a write pointer and not status. `0x50–0xA9` is just a mid-scale ADC swing. I had the same misreading baked into our firmware as a "buffer-valid flag (0x01 = complete)"; our own status line has been printing it as `0x7C`/`0x79` for weeks, which is mid-scale, and never `0x01`. Fixed on `main` tonight — we were shifted one byte late and one sample short on every frame.

For your image that means: if `debugclk_hw_top` emits three marker bytes before the samples, its framing is one byte offset from stock's, which is fine as long as we both know it. The ramp will make the offset self-evident on the first read, so this doesn't cost you anything — I just don't want us to "discover" a one-byte shift and attribute it to the mux.

Also new: **opcode `0x0A`**, which nothing in our project had ever sent because it is only reachable from op `0x09`'s handler. On bench unit #1 it returns a stable `0x89` at byte[2] while `0x09` returns `0x00` — so the pair reads `0x0089`. Stable across repeated reads, so not a free-running counter. If your netlist has a view of what backs that register I'd be glad to hear it.

**Where I actually need the debugclk image, and why it's now the critical path.** Our second channel does not work, and tonight bounded it to the fabric:

- Stock's `0x04` and `0x05` handlers are **byte-identical SPI transactions** — same opcode, same dummy, same 1024 reads — differing only in destination half of the 2048-byte buffer. There is no channel-select command, no re-arm, and nothing between them; the op-04 completion path at `0x0803ED9C` only applies gain/offset cal in place. So our read protocol already matches stock exactly.
- Two-tone rig (100 Hz into the CH1 jack, 250 Hz into CH2, DFT bin scan): **both `0x04` and `0x05` carry only the CH1 tone.** The CH2 tone is absent from both buffers. Visually confirmed too — both traces on the LCD are the same waveform.
- Swept and negative: reg `0x06` and `0x07` over 0..3 (`0x06` was the strong candidate, since stock uses `state[0x16]` as a channel index), plus `0x02` and `0x09` earlier.
- USART2 negative as well. Our coldtrace builds leave it electrically dark, so I brought it up live — `BAUDR` was already `0x30D4`, byte-identical to stock's — and sent the documented channel-config words `0x0001` type 0/1, `0x000B`, `0x0011`. No change to either buffer, and **no echo frame ever came back**; the part does not acknowledge scope commands on that wire at all.
- The load-bearing clue: **CH1's attenuator bank moves BOTH buffers, and CH2's bank moves NEITHER.** Both readout paths appear to be fed from the CH1 analog node.
- The hardware is fine — stock firmware on the same unit, same probes, renders both traces from an anti-phase pair.

So: CH1 = free-running ramp, CH2 = walking one, through the real capture/readout path, is exactly the oracle that settles it. If `0x05` returns the walking-one, the readout mux separates the channels and the fault is analog/ADC-side; if it returns the ramp, the interleave model is wrong. Nothing I can do from our firmware distinguishes those two.

Two questions while I set up the build against `25eb00b`:

1. Does the extracted netlist show anything that looks like a **channel or ADC mux select** — a fabric output driving an analog switch, or a select bit on the capture write path? Given CH1's attenuator moves both buffers, a single converter with a static mux is my leading model, and that mux control would be an FPGA pin we've never identified.
2. In your image, is the CH1/CH2 lane split decided by the same register bit the arm cone touches, or is it hard-wired in `debugclk_hw_top`? If it's hard-wired, the image tells us the mux works but not which bit drives it — still worth running, just want to know what I'm buying.

Build inputs pinned as you listed; I'll report the `.fs` sha256 and compare the chipdb msgpack first if it diverges.

DavidClawson:
**`debugclk_hw_top` reproduced byte-exact on the first attempt.**

```
9c465d72a5cd5866b7c4b06baa87f10b2e4e3c5f965f09c1518fa08c47d10751  debugclk_hw_top.fs
```

Built against `25eb00b`, your apicula `d978cad`, `DavidClawson/nextpnr@59c8f93a`. It packs to **115,638 bytes — the same length as the stock payload** — with sync `a5c3` at `0x16` and IDCODE `0x0120681B`, so it drops straight into our existing bit-bang loader. It's committed on `main` as a `guest-debugclk` build target, along with a `.fs`→C-header converter I validated by first regenerating the *stock* payload from the V1.2.0 image at file offset `0x4AD19` and diffing it byte-for-byte against our existing header.

**One correction that will save you and anyone following your recipe some time: the chipdb sha256 is not a usable diagnostic.** `apycula/GW1N-2.msgpack.xz` isn't shipped on that branch, so I built it with `chipdb_builder` from Gowin EDA vendor data — and the builder is **non-deterministic**. Five runs from identical inputs:

| run | `PYTHONHASHSEED` | chipdb sha256 |
|---|---|---|
| 1 | unset | `4dc78f30…` |
| 2 | unset | `4b70f116…` |
| 3, 4 | `0` | `26a8ac15…` (stable) |
| 5 | unset | `070b66dd…` |

None matches your `3dc7b45d…`, and none matches the one we already had (`96c46e75…`). It's Python set/dict iteration order leaking into the msgpack serialization. The decisive part: **the `.fs` came out byte-identical from two different chipdbs** (`26a8ac15…` and `070b66dd…`). So chipdb equality is neither necessary nor informative, and telling people to check it first sends them chasing a hash that can never match. Setting `PYTHONHASHSEED=0` makes it stable, but sorting the containers at serialization time would be the real fix — that's an upstream apicula improvement worth having regardless.

Also: **yosys 0.66+89 (`05805e8b9`) reproduces your hash**, so the 0.66 `86f2ddeb` pin isn't load-bearing. One build gotcha for the recipe — without `-DAPYCULA_INSTALL_PREFIX` pointing at the venv, cmake picks the system python, `import apycula` fails, and the build dies at the `chipdb-GW1N-*.bba` step while `bbasm` still links, so you get a `bbasm` but no working `nextpnr-himbaechel`.

**Answering my own question 1, from the netlist rather than from you.** I had a second pass done over the unpacked stock bitstream, and the mux model I was chasing is dead:

- **Two genuinely independent 8-bit ADC buses.** Left pads (IOL6A/B, IOL11A/B, IOL15A/B, IOL17A/B) → `BSRAM_0`; right pads (IOR1B, IOR3A/B, IOR5A/B, IOR11A/B, IOR13B) → `BSRAM_3`. **No pad appears in both.** And the whole design has only **four driven outputs** — SO, DRDY→PC0, an ADC clock forward, and one unexplained corner output — so there is no spare pin to drive an analog mux with either.
- **It's a two-stage pipeline, and the MCU does not read the front buffers.** Backward trace from SO reaches only `BSRAM_1` (depth 10) and `BSRAM_2` (depth 9); `BSRAM_0`/`BSRAM_3` never do. `BSRAM_1.DIA ← BSRAM_0` DO and `BSRAM_2.DIA ← BSRAM_3` DO, so channel identity is preserved into stage 2. **`BSRAM_1`/`BSRAM_2` are the op-04/op-05 path** — 1024 words, exactly the 1024 samples stock keeps. That also dissolves the "rate control sits on the pair we never read" tension we'd both been carrying.
- **The arm is shared**: `BSRAM_0.CEA` and `BSRAM_3.CEA` both depend on the same flop (`R13C4_DFFE_2`), with only per-channel *done* flops distinguishing them. There is no way to arm one channel and not the other.
- A channel merge does exist, but on the **readout** — ~9 LUT4s around R15C12–R16C13 whose select inputs are reachable from the SI pad and only SI. I deliberately did **not** name a bit; the shortest SI→select path runs through an AND tree fed by ALU counters, so it looks like address/count compare mixed with command state rather than a clean opcode decode, and anything more specific would be manufactured.

Two methodology notes that may matter for your RTL, since they invalidated several of our own earlier conclusions: `gowin_unpack.parse_tile_` runs with `default=True`, so pips with an **empty fuse set are emitted as if connected** — **11,242 real-fuse pips against 26,136 zero-fuse ones**, including every `SEL` wire in the netlist. So `R14C14_SEL0 = 1'b1` and friends are default tie-offs, not logic. And the 345 `` `define RxCy_E12 RxCy_EW20 `` aliases have to be resolved or the SPI SO output looks dangling when it is actually driven by `R17C12_MUX2_LUT61`. Re-deriving on real-fuse pips only, 95% of functionally-needed LUT inputs are still covered, and two independent graph constructions agree.

**Two things for you specifically.**

1. **Your `run_enable` on QN48 pin 46 conflicts with what the stock netlist shows at that pad.** `fpga/constraints/fnirsi_2c53t_qn48.cst:10` assigns `run_enable` to pin 46 / IOR1B, but in the stock design **IOR1B is a right-bus ADC data bit** (`BSRAM_3.DIA7 ← R1C20_Q4`, `DIA16 ← R1C20_Q5`, through corner-tile IOLOGIC the unpacker doesn't emit). Separately, the apicula chipdb puts that pad at **pin 35**, not 46, and its `QFN48XF` table has only 40 entries and marks IOL15/IOR3/IOR5/IOL5 unbonded even though the stock design demonstrably uses them (IOL5B is the ADC clock output). So I'd treat every QN48 pin **number** in both our projects as unverified — the apicula pad **names** are self-consistent and land correctly for IOB5A/B and IOB18A/B. This matters practically: if that assignment is wrong, `run_enable` won't reach the fabric from PB11 and the arm in your image would look broken for reasons that have nothing to do with the design.
2. **Answering your byte-2 question the other way round from before.** Your image places markers at 0/1/2 and samples at 3..1025 — 1023 samples behind a 3-byte header. Stock discards exactly **two** bytes and keeps **1024** (op-04 handler `0x0803E5A4`, op-05 `0x0803E68C`: TX one `0xFF`, discard, then 1024 reads). Confirmed independently on a second unit by Stlkv over 193 frames — second byte always `00`, third byte continuous with the window in 96.4% of frames. So the varying `0x50–0xA9` you were reading as a possible write pointer is `sample[0]`. Worth aligning your framing to stock's 2+1024 if the RTL is meant to be a drop-in.

Context for why I want the image on hardware: our CH2 is not just wrong, it's absent. Two-tone rig, engine liveness verified inside every individual reading, all 8 reachable CH2 gain settings, 2000 mVpp, and finally the physical control of moving a known-good lead between jacks — the CH2 jack reaches neither buffer, and op04/op05 return the same waveform at the same magnitude. Your ramp-and-walking-one is exactly what separates "readout serializes one memory twice" from "the second buffer has no data", so it's next on the bench.

maksidze:
I may have identified the ADC used on the board as Analog Devices AD9288, or at least a pin-compatible device.

I traced several pins on the ADC and they match the AD9288 pinout:

pin 2 — CH1 analog input → AD9288 AINA
pin 11 — CH2 analog input → AD9288 AINB
pins 17–24 — CH2 8-bit output bus
pins 37–44 — CH1 8-bit output bus
pins 25–36 also appear to match the AD9288 pinout

изображение

This is not yet confirmed, but the pinout match looks quite strong.

Since this PR currently leaves the ADC bus unconstrained, the AD9288 pinout may be useful for tracing the two 8-bit ADC buses, clock, and control signals to the FPGA.

DavidClawson:
**Before you invest further in `debugclk_hw_top`: the question it was built to answer is settled, and the image is off the critical path.** I'd rather tell you tonight than have you spend another session on it.

The image existed to separate "the readout serialises BSRAM_1 twice" from "two converters both watch the CH1 node" — CH1 ramp, CH2 walking-one. That's exactly right as a design, and I couldn't distinguish those two by content, because both predict the same waveform at a read-time lag.

What settled it was statistics rather than a new bitstream. Two *physical* converters have different offset and gain — that's why per-device calibration exists on this platform. So: three windows per trial, middle compared against the **midpoint of the outer two** (cancelling linear drift exactly), on a **flat DC input** so no window-phase wobble can bias the mean.

| condition | mean | t |
|---|---|---|
| DC, `04 05 04` | **−2.645 codes** | **−357** |
| DC, `04 04 04` — identical-opcode control | −0.003 | −0.35 |
| 100 Hz, `05 04 05` — roles swapped | **+3.429** | +54.9 |

Same read order, same timing, same drift; control exactly zero; effect mirrors on swap. **op04 and op05 read different physical converters** — your two-DPX9B model is right, and the readout mux works.

**The actual CH2 fault was MCU-side and mundane:** `ms[0x14]` (the CH1/CH2/BOTH selector) drives **PC1/PC2 as GPIO** as well as SPI3 op `0x02`. We left those pins floating, which lands in mask 1 = CH1 routed into both converters. Driving PC2 H / PC1 L gives two independent channels, reproducibly.

**Three things that bear on your RTL:**

1. **`run_enable` on QN48 pin 46 is now doubly suspect.** Beyond the stock netlist showing IOR1B as a right-bus ADC data bit, Stlkv demonstrated on his unit that holding **PB11 LOW through config and the arm writes** still configures, still arms, and still delivers frames with data (`b11_arm=0`, `A=8003F460`). So PB11 is not a run line on that hardware, and our own firmware now treats it as a CH2 relay.
2. **`0x0804BE74` is the LCD redraw dispatch table**, not an FPGA command table. Anything in your docs derived from "cmd `0x01` = configure channel, type 0/1" is built on a redraw ID.
3. **A channel-select bit exists off-fabric.** Since the mask is MCU GPIO into the analog path rather than an SPI register, a reconstruction that models channel selection as fabric state would be modelling something the real design does in hardware outside the FPGA.

**On the chipdb hash**, restating since it's still in the recipe: `apycula`'s `chipdb_builder` is non-deterministic — five runs from identical inputs gave five hashes, none matching yours. The `.fs` came out **byte-identical from two different chipdbs**, so chipdb equality is neither necessary nor informative. `PYTHONHASHSEED=0` stabilises it; sorting containers at serialisation would be the real upstream fix.

None of this makes the image useless — it still validates addressing, framing and re-arm through the real capture path, and I'd happily run it if you want that confirmed. It just isn't blocking anything now, so please don't prioritise it on our account.

Full write-ups pushed: `docs/experiments/2026-08-17-01-two-converters.md` for the paired design, and `docs/devlog/2026-08-17-two-channels-and-the-meter.md` for the narrative.

Issue #2, opened 2026-04-04
Can you make instructions for Linux? The source code is missing files


michkovitalik:
Can you make instructions for Linux?
The source code is missing files

![Image](https://github.com/user-attachments/assets/182b6964-a841-438f-9075-34c038e637bc)

Issue #4, opened 2026-04-04
I can't create this file.


michkovitalik:
I can't create this file.

![Image](https://github.com/user-attachments/assets/e84f81bc-e0ae-423b-b33a-4ce295d52e77)

michkovitalik:
[teminal.tar.gz](https://github.com/user-attachments/files/26486114/teminal.tar.gz)

michkovitalik:
<img width="1624" height="1166" alt="Image" src="https://github.com/user-attachments/assets/eaec1f93-ba29-488b-bf63-f1fa48d287a8" />

[dfu.bmp](https://github.com/user-attachments/files/26486392/dfu.bmp)

michkovitalik:
windows11 ok

[dfu.bmp](https://github.com/user-attachments/files/26486429/dfu.bmp)

michkovitalik:
<img width="1634" height="1059" alt="Image" src="https://github.com/user-attachments/assets/28b637e9-ab07-49ab-b10c-b16ab6d374d7" />

michkovitalik:
<img width="1512" height="1553" alt="Image" src="https://github.com/user-attachments/assets/e86c05b9-5277-42a9-a5d5-a6719e974c41" />

michkovitalik:
windows11 ?

![Image](https://github.com/user-attachments/assets/001fe120-e188-4b8d-a419-2e38805666b1)

michkovitalik:
I solved it with a program artery isp programmer

michkovitalik:
![Image](https://github.com/user-attachments/assets/c34fb1c5-d5ad-4376-b541-a38f8000e2dd)

DavidClawson:
Hey @michkovitalik — this is **incredibly** helpful, thank you. I've been writing the Windows flashing docs essentially blind because I only have a Mac to test on, and your screenshot is the first time I've actually seen the Artery ISP Programmer interface. The two things you circled (EOPB0(SRAM) → 224KB SRAM, then **Apply to device**) are exactly the workflow I needed to document — it's dramatically simpler than the Linux `dfu-util` command for a first-timer.

**Would you be willing to let me include this screenshot in the project documentation, with credit to you?** It would go into `docs/dfu_mode_guide.md` as a proper Windows walkthrough section, and it would save every future Windows user a lot of confusion. Happy to link back to your GitHub profile as the contributor.

If you're up for it, two quick follow-up questions to round out the walkthrough:

[...]

PR #29, opened 2026-08-21
shell: USB-only firmware updates — stage over CDC, cache on the W25Q, install with a clean reset

Stlkv:
**This firmware can now update itself over the USB cable it is already plugged into — no disconnecting, no buttons, no MENU+Power dance.** One command on the host streams a new image over the CDC shell, the device verifies it at rest and installs it with a clean reset:

```
python3 scripts/cdc_flash.py firmware/build/firmware.bin
```

Three things to know up front:

1. **Stock is always one button away.** Nothing in this path CAN write below `0x08007000` — the geometry is compile-time constant, so the factory IAP bootloader is physically untouchable. MENU+Power keeps working as the recovery and back-to-stock path after any outcome, including power loss mid-install.
2. **The changes to existing code are small.** Against current `main`: 278 insertions / 22 deletions in existing files, 1 033 lines of new modules, plus 410 lines of experiment writeups. The X-Y RAM-ceiling resolution that used to be listed here is no longer part of this diff — it landed on `main` as `b2723a4`.
3. **As a bonus**, the two 1 MB W25Q cache slots this introduces use a manifest format shared with the 2C23T port, so a unit can keep BOTH firmwares cached and flip between them in seconds with `fwswap a|b` — no host transfer at all. Unit #2 has been switching back and forth on one cable for days; the factory-cal page stayed byte-identical (CRC `DFD5381C`) across every hop. One note for the swap path: this firmware's boot assumes a COLD FPGA, so after swapping INTO it from another firmware, do one ordinary power-cycle (unplug) if capture comes up dark — the config port read on a warm FPGA is the known desync (your Exp L note), not a new defect.

### Using the flash script

```
python3 scripts/cdc_flash.py <image.bin> [--port /dev/...] [--slot a|b] [--stage-only]
```

1. Build (`make guest-coldtrace`) or take any `0x08007000`-linked image up to 740 KB.
2. Run the script against the device's CDC port (it auto-detects the first `usbmodem`/`ttyACM` if `--port` is omitted). It sends `fwload <size> <crc32> <slot>`, streams the raw bytes with progress, and waits for the device's verdict line: `fwload: STAGED ...` means the slot re-read clean against the announced CRC and carries an app-shaped vector table; anything else aborts with the reason printed, and the running firmware is untouched — resend at will.
3. Unless `--stage-only` was given, the script then issues `fwapply`: the device re-verifies the ENTIRE slot at rest, installs from RAM with read-back verification, and system-resets into the new image. The CDC port drops and re-enumerates — that is success, not a crash. Keep USB attached: the cable carries the power rail through the reset.
4. `--stage-only` fills a cache slot without installing (useful for pre-loading the second firmware); `fwswap a|b` in the shell installs a cached slot any time later, and `fwstat` shows the slot table.

### What was touched (deliberately little)

| file | change |
|---|---|
| `usb_debug.c` | +215 — three shell commands, RX rerouting to the loader while a transfer runs, verdict printing, drain handling |
| `Makefile` | +17 — build the new module and its host test, the test target declared `.PHONY` on its own line so the shared list stays untouched |
| `flash_regions.c/h` | +12 — one new region: `fwcache` 0xD00000..0xF00000, carved from uservol's read-only tail |
| `tests/test_config_persist.c` | +30/−19 — its hostile-table fixture derived from `flash_region_table[]` instead of hand-listed ids (see finding 4) |
| `tests/test_flash_regions.c` | +4 — the new region in the table checks |
| `fw_loader.c/h` | new, 714 — the state machine and the RAM installer |
| `tests/test_fw_loader.c` | new, 195 — host tests for every gate that stands before an erase |
| `scripts/cdc_flash.py` | new, 124 — the host half |
| `docs/experiments/` | new, 410 — four writeups: EXP-25, EXP-201, EXP-202, EXP-203 |

### Mechanism

1. `fwload <size> <crc32> [a|b]` reroutes CDC RX into a pure state machine (`fw_loader.c` — prints nothing, knows nothing about USB, host-testable). Bytes stream into a W25Q cache slot through the `flash_regions` layer, lazy sector erase, write-verify per chunk. On completion the slot is re-read AT REST, CRC-checked, its vector table shape-checked, and only then its manifest is written — a torn transfer can never look installable, to this firmware or to the port's. A per-page FWDGT feed keeps an armed watchdog alive through the install (your must-fix; `fw_loader.c:458`).
2. `fwapply` / `fwswap a|b` re-verify the ENTIRE slot at the moment of truth, then a RAM-resident installer reads the W25Q over SPI2 directly (polled, self-contained — the flash-resident driver is unreachable while the app slot is erased under it), programs the slot page by page **routing between the two internal-flash banks by address**, verifies by read-back, and ends in SYSRESETREQ — a clean boot of the new image.
3. Run installs on USB power: PC9 drops during the reset and the cable carries the rail — guaranteed by construction for anything this PR does.

### Verified on hardware

| claim | evidence |
|---|---|
| host streams an image over CDC, verified at rest, installed via reset | measured on both units, both firmwares; latest run 604 336 B / crc `BB211CBE`, 633 s uptime after install with no reset |
| cross-firmware cache: either firmware installs what the other staged | measured in both directions in one session — `fwswap b` into this firmware, `fwswap a` back into the 2C23T port |
| a torn transfer can never look installable | host tests plus manifest-written-last; the silence-timeout and drain paths exercised on hardware on both firmwares (EXP-202) |
| the FWDGT feed survives a real install | measured under a live 3.0 s watchdog, armed by register readback, ~16 s install window, `wdtrstf` clear afterwards (EXP-201) |
| nothing can write below `0x08007000`; MENU+Power always recovers | compile-time geometry; recovery exercised more than once during this PR's history |
| the factory-cal page is never touched | CRC `DFD5381C` re-read after every install; `cal backup` on unit #2 reports `match : yes` against the same value (EXP-203) |

**Not verified, stated plainly:** the measurement side has only ever run its refusal branches on unit #2 — there was no signal generator on that bench, so rise/fall, graticule scale and X-Y figures are untested there. The malloc-failure branch of the X-Y snapshot buffers has not been exercised either. Neither is on this PR's path, but both are claims I am not making.

### Five findings from the bench, part of this PR's history

1. **A cross-firmware jump half-bricks.** The first revision jumped (VTOR+MSP+bx) into the new image; USB stayed up on interrupts, main loop dead, screen dark. A jump hands the new image the old firmware's RTOS/peripheral posture, which no init sequence owes correctness under. Hence the system reset. (Recovered via NRST; nothing was lost.)
2. **The bank boundary at `0x08080000`.** This image is ~597 KB and crosses into flash bank 1, where bank-0-routed register writes are silently dropped — an installer that ignores this dies at verify on any image over ~468 KB. Ours routes per page; worth knowing for any future flash writer.
3. **`flash_regions` was never bound on target.** The layer is host-tested and audited, but nothing in the firmware called `flash_regions_bind_w25q()`, so on hardware every region op returned `NOT_INIT`. The loader binds lazily; `settings_store_init()` now binds at boot as well.
4. **The region ids must stay in address order** — the table self-check requires ascending starts and fails the whole table closed otherwise (good design; it caught our own first attempt). It also caught us a second time, in a test fixture: a hostile table listing four ids by designated initializer with a literal count of 4 grew a hole when `FLASH_REGION_FWCACHE` was inserted at id 2. Fixture fixed, not the check.
5. **The W25Q "write defect" on unit #2 was never the chip.** This PR's `fwload` died at 512 B there for weeks while `fwswap` worked and reads were flawless. A side-by-side of the two firmwares' writers on the same clone pointed at the driver rather than the protocol, and DavidClawson found it: a tSHSL (CS deselect time) violation in `flash_fs`, fixed in `479d120`. Replicated on unit #2 with nothing but the merge — `flash wtest` passes and a 598 948 B stream stages clean (EXP-25).

### Follow-ups this enables

1. ~~A boot-time `flash_regions_bind_w25q()`~~ — done: `settings_store_init()` binds at boot. Confirmed on hardware, not just read: unit #2's `cal backup` succeeded in a session where no `fwload` had run, so nothing else could have bound the layer.
2. The cal-BIN transfer discussed in #18 — one raw-byte CDC channel, several consumers.
3. A shared flash-map document: app-slot ceiling `0x080C0000` (the port keeps its FPGA bitstream store there), fwcache at W25Q `0xD00000..0xF00000` — both firmwares honor these today; a doc would keep it that way.


DavidClawson:
@Stlkv — this is excellent, and it's the good kind of paranoid. The manifest-written-last + full at-rest re-verification, the system-reset-never-jump (with the 08-22 half-brick recorded right there in the comment so nobody re-learns it), the per-page bank routing across `0x08080000`, and the RAM-resident installer carrying its own SPI2 driver because the flash-resident one is unreachable mid-erase — all of that is exactly right. I read the installer closely and no libc/libgcc call sneaks into the interrupt-disabled window (no div/mod, no `memcpy` reaching into erased flash). And "nothing can write below `0x08007000`" means recovery is always one button away, which is the property that lets me actually trust this on the bench.

**One must-fix before I flash it here, and it's specific to this firmware, not yours.** `fwl_ram_install()` (`fw_loader.c:419`) does `cpsid i` and then erases + programs the whole ~595 KB app slot in one stretch with no watchdog refresh. Our `guest`/`coldtrace` builds run the FWDGT with a **hard 3.0 s timeout** (`watchdog.c:37`), normally fed from the 500 ms health task — which stops dead the moment interrupts go off. The install is comfortably 10–20 s (erase alone is ~298 pages), so the watchdog fires partway through the **erase**, and since the vector page at `0x08007000` is erased first, the reset lands on an unbootable slot. Recoverable via MENU+Power, but the feature fails looking like a brick.

I'm fairly sure this is why it round-tripped all night on your unit and hasn't been seen here yet — you're on the 2C23T port, which has a different watchdog posture. The fix is one RAM-safe store at the top of the per-page loop (`fw_loader.c:443`):

```c
*(volatile uint32_t *)0x40003000u = 0x0000AAAAu; /* WDT reload — MUST stay inside the loop */
```

One feed per page keeps us far under 3 s. Happy to push that myself if you'd rather, but it's your PR — your call.

**One thing to sanity-check:** inserting `FLASH_REGION_FWCACHE` mid-enum shifts the numeric ids of `USER_CAL`/`SETTINGS`/`MODULES`. Symbolic use is fine, but is any region id ever *persisted* (an append-log header) or handed to a host tool by number? If not, non-issue — and the address-ascending constraint you documented is real and correct.

**On your W25Q question:** I'll run `flash wtest` on unit #1 and report back — that decides whether the streaming half works here or whether I'm on `fwswap` like you. Either way the swap path + shared manifest is a real win; I did the MENU+Power dance about six times in one session this week, so this lands right where it hurts.

With the watchdog feed in, this is merge-after-one-small-change. Thank you — this is a lot of careful work.

Stlkv:
No rush on the review — just a heads-up that the branch grew by two commits, and a report from an odd test rig.

The 2C23T port for this board now speaks this PR's protocol verbatim, so `cdc_flash.py` here drives that firmware and its `cdc_bench.py` drives this one. An evening of running one protocol from both ends shook out three bugs, all in code this PR introduced (`git log -S` puts each line in my own commits), two of them in failure paths rather than the happy path. Fixed in `5f22dd0`. I have marked what is measured, what is read off the code, and what is a judgement call, because they are not equally solid.

#### 1. An aborted transfer left the rest of the image to the line editor

*Code fact:* once `fw_loader` fails mid-stream, `fw_loader_active()` is false, so the next RX takes the `else` branch into `shell_feed()`. *Not measured:* what that looked like on the old build — `cdc_flash.py` aborts on the first `ERROR` chunk, so seeing it needs a host that ignores `ERROR`, and I never ran one against the old code. The fix swallows exactly what the aborted image still owed, then gives up after roughly the loader's own silence window (same constant, slightly different clock: my counter ages only on idle iterations). *Measured*, naive host streaming all 131 052 B: the device emitted **117 bytes total** — one status line at the real failure, then `aborted image drained; shell is listening again` — and `fwstat` answered normally. One failure mode only; a packet carrying both the tail of a dead image and the start of a real command is handled by construction, untested.

##### The trade-off, named rather than hidden

*Judgement call, not a measurement:* while the drain is armed it also eats a command a human types in that window. I could not find a rule separating the two that survives a binary — image bytes carry CR and LF constantly — so I took the bounded swallow. If you see a better rule, I would rather have yours.

#### 2. `fwload <size> <crc> c` armed the intake on the default slot

A third argument that is neither `a` nor `b` fell back silently and armed anyway. *Measured, on the port* (same fallback in its parser): the shell went into raw mode while I was still reading the reply, my next command line went into the image, and the port stayed mute until a power cycle. *Code fact, not a second incident:* `cmd_fwload` here had the identical fallback. Now refused — measured that the state stays `IDLE` and the shell answers. Minimal change: a valid letter followed by junk is still accepted, as before; say the word if you want that strict too.

#### 3. `cdc_flash.py` missed a verdict it had already printed

*Measured* on a 604 132 B image: `fwload: STAGED slot=b ... crc=9bed9bde` appeared inside the output the write loop was draining, the post-loop `read_until` found nothing, the tool exited "staging did not verify" — and `fwstat` showed the slot correctly staged. *Reading, not separately measured:* `STAGED` was scanned for only after the loop, and at this size the device finishes before the loop writes its last chunk. Streamed output is now kept in a bounded buffer and checked too; same image, same command, now reports STAGED.

#### Two of your decisions I only appreciated by getting them wrong

The RX-silence timeout is not decoration — my port had none, and an operator typing into an armed intake wedged it until power cycle. I copied yours. I then tried to be "more honest" than `fw_loader_begin()` by erasing the manifest at `fwload` time, which is worse: *code fact*, your erase watermark starts at the slot base, so the manifest dies with the first committed data and a transfer that dies before any byte leaves the cached image installable. *Verified on my reimplementation of your semantics, not on this firmware*: armed then left silent, the port's slot still read `131052/2BCD5FC3`, where my erase-at-begin version had wiped it. Reverted to yours. Thanks for both.

#### One trap, now a line in the script's docstring

`make` produces the HID-bootloader flavour — `.isr_vector` at 0x08004000 — so `objcopy` emits a file based there; staged into a cache slot on a unit with the FACTORY bootloader it lands 0x3000 low and the device does not return (measured: staged, installed, dark). My mistake, not a defect, but worth a line because **neither gate catches it**: at offset 0 the file holds a real vector table (SP `20037FE0`, PC `08027CB5`), so `fwl_finish`'s SRAM-SP and in-slot-PC checks pass honestly, and my cache checks CRC only. I see no cheap check that would catch it — an opinion, not a proof; if you see one, I would rather have the check. `make guest` is the flavour for such a unit. Recovery was what the design promises: MENU+Power and the stock IAP.

#### Verified, and not verified

`make test-fw-loader` passes but covers the loader's gates, none of these three paths — `usb_debug.c` and the host script have no tests, so all three rest on the bench runs above. The second commit only publishes the silence-timeout constant so both users read one symbol; the bench runs predate that rename, and since this firmware embeds `__DATE__`/`__TIME__` I cannot even offer a byte-identical image as evidence. Cross-firmware state agreed: both firmwares read the cache identically (`A=131052/2BCD5FC3`, `B=604132/9BED9BDE`), this firmware installed from slot B by the port, the port reinstalled from slot A by this firmware's `fwswap`. Shell level only — I did not re-check scope or meter telemetry on this build.

#### Still open — I would value your read

`fwload` into this unit's W25Q still fails on your firmware (`err=w25q erase/write/read` at 512 B, same signature as on the 22nd) while `fwswap` from a cached slot works and reads are flawless. I have been calling that a write-path defect, but the honest version is narrower: the port's own writer succeeds on this same chip (Zbit ZB25VQ128 clone, id 5E17) where this one does not, which points at the driver or its bus setup rather than the protocol, without isolating which. I can dump both write paths' register sequences side by side on the same chip if that would help.

DavidClawson:
**Found it — and you called it exactly right.** "The driver or its bus setup rather than the protocol" was precisely where it lived. The root cause is a **tSHSL (CS-high / deselect-time) violation** in `flash_fs`, fixed in `479d120` (pushed to `main`).

### What it was

Every W25Q write path is `WREN` then the command as **two back-to-back SPI transactions**. Between them the CS deassert→reassert took only a few instructions — below the flash's minimum CS-high time. The chip never registered the deselect, treated the second command (the `RDSR` inside `raw_write_enable`) as a continuation of the WREN frame, and left MISO **undriven**. So `raw_write_enable` read a genuine `0xFF`, decided the bus was dead, and aborted every erase and program before it started.

### How it was cornered

`flash wtest` failed on unit #1 with your exact signature. Then a step-resolved diagnostic that **inlined** `WREN → erase → program` passed every time, while the public `flash_fs_raw_sector_erase`/`write_block` failed every time *on the same address in the same session*. Ruled out, one instrumented build each: chip alive (JEDEC `EF4018`, reads perfect), not write-protected (BP/WPS clear, WEL latches), not the malloc (14 KB heap free, succeeds), not a PB12/SPI2 clobber (`force_reinit` changed nothing), no transfer timeout. The one difference left between the working inline path and the failing wrapper was a **function call's worth of CS-high time** between the two transactions — i.e. tSHSL. Adding ~1 µs of CS-high after every deassert fixed it.

### Why it matched your cross-chip result so well

It reproduced on unit #1's **genuine Winbond** and your **Zbit ZB25VQ128 clone** because it was never the silicon — it was our driver's timing, exactly as your side-by-side pointed. Same one line also explains `settings writes ok: 0 ever` and the cal-backup path having nowhere to write.

`flash wtest` now passes on unit #1 (page-program in place, erase+RMW, sector restored). **So `fwload` streaming should now work here, not just `fwswap`** — I'll confirm against your `cdc_flash.py` and report the round-trip. If you re-pull `main` onto your unit, I'd expect your `err=w25q erase/write/read` at 512 B to clear too. (Thank you for the offer to dump both writers' register sequences — the fix landed before I needed to take you up on it, but that instinct is what made this findable.)

### On the PR itself

Your three fixes in `5f22dd0` all read right — the bounded drain of an aborted image, the strict slot-letter refusal, and scanning the streamed buffer for the `STAGED` verdict (the large-image race). The named trade-off on the drain (it eats a human keystroke typed into an armed intake) is the honest call; I don't have a better rule either — image bytes carry CR/LF constantly, so there's no clean separator. Leave it as you have it. And the "`make` produces the HID-bootloader flavour, lands 0x3000 low, both gates pass honestly" trap is a real one worth the docstring line — neither shape check can catch a validly-shaped image linked for the wrong base.

**One thing still outstanding before this merges for our units** (separate from the W25Q fix): the install step. `fwl_ram_install` erases+programs the app slot with interrupts off for ~20 s, and our `guest`/`coldtrace` builds run the FWDGT at a hard 3.0 s timeout (`watchdog.c:37`) fed only from the halted 500 ms health task — so the watchdog trips mid-erase. That's the IWDG-feed fix I flagged up top (one store, `0xAAAA` to `0x40003000` at the top of the per-page loop). With the W25Q write path fixed *and* that feed added, the whole `fwload → fwapply` path should run cold on a stock-bootloader unit. Happy to add the feed to your branch if you'd rather I did.

Stlkv:
@DavidClawson — the tSHSL find is a beautiful piece of debugging, and I have the strongest thing a second bench can offer in return: **it replicates on unit #2, on the Zbit clone, with no change but the merge.** Full session below — plus your watchdog feed already in and proven under a live FWDGT, one RAM-ceiling regression the merge surfaced (with a proposed fix you can veto), and answers to both open questions from your review.

## 1. `479d120` cures unit #2 — measured tonight

Merged your `main` into this branch (`e9bb064`), built unmodified `guest-coldtrace` (598 948 B, crc `D5A4C58C`), staged it via the port's `fwswap`-safe path, and ran the previously-failing paths on THIS firmware's own driver (record: `docs/experiments/2026-08-25-25-tshsl-fix-replicated-unit2.md`):

1. `flash wtest 0xFA0000 CONFIRM`: **all three phases PASS** — first wtest pass ever recorded on this unit.
2. The end-to-end half that died at byte 512 on 08-22 and 08-24: `cdc_flash.py <598 948 B image> --slot b --stage-only` → **`fwload: STAGED slot=b 598948/598948 crc=D5A4C58C err=none`**, 67.3 s, full at-rest CRC re-read clean. ~147 sectors of erase+program on the chip that "had a write defect". It never did — your diagnosis was exact.
3. Honestly labeled: `settings` still reads `writes ok: 0` here, but that is **no data, not a failure** — no save was attempted this boot (saves fire on button press / orderly power-off; the session was shell-only). The primitive underneath is covered by wtest.

**So the PR's "Known limit" section is obsolete: both halves of the update path now work on both firmwares, on both bench units.** I'll edit the description accordingly.

## 2. Your must-fix landed before your comment — and it has now run under a live FWDGT

Our messages crossed: the feed went in yesterday (`1136141`, your line verbatim, `fw_loader.c:458`). Rather than leave it code-reviewed-only, I armed the dog on unit #2 and made it carry a real install (`docs/experiments/2026-08-25-23-fwdgt-feed-under-live-watchdog.md`):

1. Bench-only build: `guest-coldtrace` with the `main.c` watchdog guard bypassed locally (one line, committed nowhere). FWDGT confirmed armed **by register readback**: divider 4 (DIV_64), reload `0x753` = 1875 → your 3.0 s.
2. `fwswap b` — the armed firmware reinstalling **itself**, 597 KB ≈ 292 pages: install window **~16 s, five-plus watchdog periods**, interrupts off throughout. Boots back clean.
3. `CRM_CTRLSTS = 0x14000000` after boot — `swrstf` + `nrstf`, **bit 29 (`wdtrstf`) clear: the dog never fired.** Second data point: the 131 KB port install, ~4.5 s window.
4. Stated weakness: the counterfactual (feed removed → brick) was derived from the FWDGT being a hardware counter `cpsid i` cannot pause — not demonstrated. One MENU+Power recovery buys the symmetric data point if you want it.

**One correction to the "why did no bench see it" hypothesis, and it applies to your bench too.** Code facts, on your `main`: `main.c:1067` gates `watchdog_init()` behind `#ifndef GUEST_BUILD`, that call is the only FWDGT arm site in the linked sources (`src/attic/` is not in `C_SOURCES`), and **every `guest*`/`coldtrace*` target defines `GUEST_BUILD`** — including `guest-coldtrace` (`Makefile:833`) and the v0.3.0 release flavour. So on coldtrace builds the failure was never reachable on unit #1 either; it is live exactly on plain-linked (`make`, `0x08004000`, HID-bootloader) builds. Which flavour does your bench flash? Either way the feed stays: one MMIO store per page, a no-op on an unarmed dog, and EXP-23 shows it holds when the dog is real.

## 3. The merge surfaced a RAM-ceiling regression — proposed fix inside, yours to veto

Merging current `main` into this branch stopped linking, and the numbers say it is not just this PR (`6880b96`):

| build | result |
|---|---|
| `make guest`, main `df96059` **alone** | **352 B over RAM** |
| `make guest`, main + this PR | 1 576 B over |
| `make guest-coldtrace`, main + this PR | 984 B over |
| `make guest-coldtrace`, main alone | links (why no bench saw it) |

The X-Y view's two 1 KB static snapshot buffers did it; this PR's 1 223 B (680 B RAM-resident installer + 512 B intake buffer + state) merely made it unmissable. `6880b96` carries the minimal resolution so every flavour builds: the snapshot pair moves to the FreeRTOS heap, allocated once on the first X-Y frame with data (`scope_ui.c:1570`); allocation failure degrades to "no snapshot this frame", never a null deref. It's your feature — happy to swap this for whatever resolution you prefer.

## 4. Your region-id question: nothing persists or exports the number

All code facts, on this branch: the append-log record header is `magic 0xA5C3 / u16 len / u32 crc32` (`flash_regions.c:528`) — no region id in it; the single append user binds symbolically (`config.c:31`, `#define CONFIG_REGION FLASH_REGION_SETTINGS`; `settings_store.c` reaches flash only through it); no shell command or host tool speaks region numbers. The one place an id becomes a number, `fw_loader.c:171`, derives it from the table pointer at the moment of use and never stores it. Mid-enum insert is safe; the address-ascending self-check pins every id to its range.

## 5. Two more of my own bugs, fixed and bench-run — and the safety invariant re-checked

Both in the drain logic, commit `1136141`, both exercised on hardware on both firmwares (`docs/experiments/2026-08-25-24-silent-death-drain-both-firmwares.md`):

1. A transfer killed by the RX-silence timeout did not arm the drain — a host that stalled >3 s and *resumed* fed the rest of the image to the line editor. The poll path now arms it (`usb_debug.c:7082`). Measured: stall at 256/597 464 → verdict line; resumed 64 KiB → **zero bytes of shell output**; abandoned drain → `stopped waiting`, shell answers. Same three-way pass on the 2C23T port (its commit `008e36e`).
2. The feed-path owed figure over-counted by the tail of the failing packet — up to 63 bytes of the operator's next keystrokes eaten. Now counted from bytes actually sent (`usb_debug.c:7034`). Labeled honestly: host-tested and code-reviewed, NOT bench-exercised — both our host tools write in 64-multiple chunks; it needs a host with odd write granularity.

And after tonight's ~8 installs and streams, the invariant the PR description leads with was re-measured, not assumed: **the factory-cal page read back byte-identical, crc `DFD5381C`** — same value as before any of this landed.

## 6. State of the goals, one table

| PR claim | status tonight |
|---|---|
| host streams image over CDC, verified at rest, installed via reset | measured, both firmwares, both units |
| nothing can write below `0x08007000`; MENU+Power always recovers | code fact (compile-time geometry); recovery exercised twice this week |
| two W25Q cache slots, manifest shared with the 2C23T port, `fwswap a\|b` | measured again tonight, cross-firmware both directions |
| `--stage-only`, `fwstat`, verdict lines | exercised tonight, three stagings |
| torn transfer can never look installable | host tests + manifest-written-last; drain fixes close the shell-side holes |
| cal page survives installs (crc `DFD5381C`) | re-measured tonight after ~8 installs: identical |
| **"Known limit": unit #2 W25Q write defect blocks streaming** | **obsolete — your `479d120`; description edit coming** |
| watchdog feed (your must-fix) | in (`fw_loader.c:458`) and proven under a live FWDGT (EXP-23) |

With the feed in, the tSHSL fix on `main`, and the RAM resolution in `6880b96`, I believe every flavour builds and every claimed behaviour is measured. The one open thing on my side is the odd-granularity host for point 5.2 — cosmetic, not blocking.

DavidClawson:
@Stlkv — the unit-#2 replication is the result that matters most here, and it's the strongest possible close: **cross-unit *and* cross-silicon** (Winbond on #1, Zbit clone on #2), same behaviour, cured by nothing but the merge. That retires "it's the chip" permanently — it was always the driver's inter-command CS timing. `flash wtest` passing on a unit that had never passed it, plus the 598,948 B `--stage-only` round-trip at rest, is exactly the evidence I couldn't produce from one bench. Thank you for turning it around the same night.

**On the watchdog feed — our messages crossed (twice now, it turns out).** I'd independently prepped the identical one-line feed on my side before your comment posted, so I'm glad yours landed first *and* came with something I couldn't give it: a run under a live FWDGT. Register-readback of DIV_64 / reload `0x753` → 3.0 s, a ~16 s self-reinstall with interrupts off, and `CRM_CTRLSTS` bit 29 clear afterward — that's the proof, not the code review. Mine is redundant; I'm dropping it.

And your **`GUEST_BUILD` catch is the real insight**, so let me answer directly: **my bench flashes `guest-coldtrace`** — which defines `GUEST_BUILD`, so `watchdog_init()` is `#ifndef`-gated out and the dog was **never armed on unit #1 either**. That fully explains why neither bench saw the brick: the failure is live only on plain-linked `make` / `0x08004000` / HID-bootloader builds. Which is precisely the point in favour of keeping the feed — **that HID path is the end-user update path** (closed-case, USB-only), so the one flavour a user actually runs is the one where an unfed dog would trip mid-install. One MMIO store per page, a no-op when the dog is unarmed, live insurance when it isn't. Keep it.

**On the RAM regression — `6880b96` accepted, and independently confirmed.** I bisected the same ceiling this session: `make guest` on `df96059` (main, *before* your PR) is **352 B over**, byte-for-byte your number. So this is a **main-branch regression your PR merely made unmissable** — the X-Y snapshot pair, my feature, not yours. Moving the two buffers to the FreeRTOS heap with allocate-on-first-XY-frame and degrade-to-no-snapshot on failure is exactly the right resolution (2 KB out of a 32 KB heap, no null-deref path). I'll make sure it lands on `main` so every flavour links, not just this branch — the fix belongs there regardless of the PR.

**On the region-id question (#4):** thank you for chasing it to ground — "the id is derived from the table pointer at point of use and never stored" is the answer I needed, and it retroactively validates something on my side: I added a factory-cal backup feature this week that inserts a new region *before* `TAIL`, and your finding is what makes that mid-enum insert safe. Related, and a nice property of your install path: the factory-cal page reading back byte-identical (crc `DFD5381C`) after ~8 installs means your path genuinely never touches `0x08006000`. That's complementary to the backup work — you preserve it, I mirror it off-chip against the region *our own app* overwrites.

The two drain fixes (silence-timeout not arming the drain; the owed-count over-count) are good catches and the honest "5.2 host-tested, not bench-exercised — needs odd write granularity" label is the right call; that one's genuinely cosmetic.

Net: with the feed in, the tSHSL fix on `main`, and `6880b96`, I agree — every flavour builds and every claimed behaviour is measured. The "Known limit" removal is earned. This has been a model of how to close a PR: two benches, two chips, experiment files for each claim. 🙏

Stlkv:
@DavidClawson — merged `main` twice today (`ea932d7` at `857802e`, `370ab09` at `c84646b`), so this branch is level and no longer conflicting; both conflicts were one-liners, an include block and the `.PHONY` line. Checked before merging: your `b2723a4` and this branch's `6880b96` are line-identical, and the merged `scope_ui.c` has exactly one allocation site — nothing duplicated. Thank you for taking it to `main`, that is where it belonged.

Then I put the merged tree on unit #2. Three things came out of it, one of them yours.

#### Your cal-backup, confirmed on the second bench and the second chip

Measured tonight on `guest-coldtrace` at `ea932d7`, one session, Zbit ZB25VQ128 clone:

| step | reading |
|---|---|
| `cal status` before | `live @0x08006000 programmed crc=0xDFD5381C` / `backup: no backup (bad magic)` |
| `cal backup` | `ok` |
| `cal status` after | `backup: valid v1 payload_crc=0xDFD5381C src=0x08006000` / `match : yes` |
| `flash read 0xFFD000 32` | `42 43 41 4C 01 00 00 10 00 60 00 08 1C 38 D5 DF` — `BCAL` / v1 / len `0x1000` / src `0x08006000` / crc `DFD5381C` |
| live page after | `0xDFD5381C`, unchanged from every earlier reading here |

The "no backup" line before is what makes the rest mean something: the region was empty at session start. **`cal restore` deliberately not run** — it writes MCU flash, so only the W25Q half is covered. That auto-restore could not have fired is a **code reading, not a measurement** (`#if defined(CAL_BACKUP_AUTO_RESTORE)`, and `cal_backup_should_auto_restore()` refuses a programmed page).

#### The X-Y heap change, with the numbers it never had

You accepted `6880b96` on the reasoning alone; here is the evidence, read over the shell against the installed image's ELF symbols.

| reading | before X-Y | after |
|---|---|---|
| `xFreeBytesRemaining` | `0x3770` = 14 192 B | `0x2F60` = 12 128 B |
| `xMinimumEverFreeBytesRemaining` | `0x3770` | `0x2F60` |
| `xy_xs` / `xy_ys` | `0` / `0` | `0x20033CC8` / `0x200340D0` |
| `ui_scope_full_draws` | 1 | 27 |

−2 064 B is exactly 2 × (1 024 + 8 B `heap_4` header), the pointers are `0x408` apart, and min-ever-free tracks that single drop with no deeper dip. The allocation is gated on `fpga_data_ready()`, so the snapshot ran on real capture data.

**Weaknesses, unprompted:** X-Y was entered by writing `scope_view` over the shell, not by the button; one entry only, so "allocated once and kept" is still the code's claim; the malloc-failure branch never ran with 14 KB free. **And no generator was on the bench**, so the whole measurement side ran refusal branches only — `fpga scope measure` prints your `rise1_smp100` / `fall1_smp100` as `-` at `pp1=3` and `k1_uV=0`, which is the behaviour I want but says nothing about the values. Graticule `autofit`, softtrig `ON`. I will do the signal half when the rig is back.

Incidental but relevant to this PR: the install path ran both directions in the same hour — the port's tool streamed 604 336 B (`STAGED`, crc `BB211CBE`), `fwswap b` installed it, 633 s up with no reset, `fwswap a` put the port back. Record: `docs/experiments/2026-08-26-203-merged-tree-boots-and-cal-backup-on-unit2.md`.

#### One catch that is `main`'s, not this PR's

With `arm-none-eabi-gcc` on `PATH`, `scripts/test_firmware_build.py` fails its warning ratchet on three `'spi3_xfer2_gapless' defined but not used` (app, emu, guest). **Reproduced on `origin/main` alone in a clean worktree**, so it is not a merge artifact: `42c831b` defines the helper unconditionally while every call site sits under `#if FPGA_HW_PRELUDE_GAPLESS`, default 0. Wrapping the definition in the same `#if` matches how that file gates its experiment-only code. I have not patched it — your file, your ratchet, and I would rather not widen this diff — but say the word and I will push it. Why neither of us saw it: with no toolchain the suite skips all five build tests, and its own skip message says that is missing coverage, not a pass. That message is what made me re-run.

#### Correcting yesterday's citations, and a numbering proposal

Two files I cited yesterday have moved, and one sentence now points at your work: where I wrote "EXP-23 shows it holds when the dog is real", EXP-23 on `main` is your config-transport runbook. Sorry — stale within hours of posting. We share one counter and you fill it faster, so mine went 20 → 23 → 26 in a day, colliding each time. From today **unit #2's writeups take a reserved 201+ block**: EXP-201 (FWDGT feed, cited yesterday as `…-23-…`), EXP-202 (silent-death drain, `…-24-…`), EXP-203 (tonight). Each file records its former numbers. **EXP-25 keeps its number** — no collision, and you answered it by name. If you would rather a different scheme, I will follow yours; I only need one that does not need a rename per day.

#### And one claim of mine that needed a fix

Yesterday I told you a mid-enum region insert is safe because nothing persists or exports the id. That still holds at runtime — but a fixture on this branch did depend on enum order, and I had not run that suite: `test_config_persist`'s hostile table listed four entries by designated initializer against a literal count of 4, so `FLASH_REGION_FWCACHE` at id 2 left a hole and the self-check refused the table. Failing since the cache landed, not since the merge (reproduced on the pre-merge commit); fixed in `c5a6c4e` by deriving the fixture from `flash_region_table[]`, with the `FLASH_REGION_ERR_READ_ONLY` assertion intact so it stays a real negative control. 22/22.

After both merges: `make guest` 611 164 B and `make guest-coldtrace` 604 336 B both link, `scripts/run_tests.py` is 13/13 suites / 95 tests, and the C tests pass. The "Known limit" section is obsolete since your tSHSL fix — editing it out and taking this PR out of draft with this comment.

Stlkv:
@DavidClawson — `main` is merged in (`55260e3` at `0b7f9d0`) and the branch is clean again. Same one-line conflict as last time, the shared `.PHONY` list, so this time I went after the cause rather than the instance (`fcb53da`): `test-fw-loader` gets its own `.PHONY` line next to its target, and the shared line is byte-identical with yours again. Checked rather than assumed — I replayed a synthetic upstream commit appending a target to that line and merged it: clean. Should stop being a papercut for both of us. The Makefile hunk is pure insertion now, +17/−0, and that is the only change to this PR's diff: 14 files, 1 721 / 22, of which 278 / 22 touch existing files. Description updated for those two numbers.

#### One thing the merge surfaced, and it is `main`'s, not this PR's

`make guest`, `make` (app) and `make emu` do not compile at `0b7f9d0`. **Measured** in a clean `origin/main` worktree with nothing of mine in it:

```
src/drivers/fpga.c:5816:18: error: 'BB_SCK' undeclared (first use in this function)
src/drivers/fpga.c:5830:9:  error: implicit declaration of function 'bb_cmd16'
make: *** [build/fpga.o] Error 1
```

**Code fact:** `fpga_edgecap_15()` (`e803915`) compiles unconditionally while `BB_*` and `bb_cmd16()` live under `#if FPGA_CONFIG_B`. Very easy to miss, and I would have — `guest-coldtrace` and `guest-bringup-bb` still build (measured), and those are exactly the two flavours we each flash. So if `make guest` fails on this branch it is inherited: **the failure profile here and on bare `origin/main` is identical**, measured, 12/13 suites and 95 tests on both, same five cases failing inside `test_firmware_build.py`.

Rather than widen this diff with your file, the fix is a separate small PR: #NN, 3 files, +38/−15 — the same `#if FPGA_CONFIG_B` on `fpga_edgecap_15()` and its prototype, `spi3 edgecap` kept in the shell table everywhere with an answer naming the flavour to flash, and `spi3_xfer2_gapless()` gated on `FPGA_HW_PRELUDE_GAPLESS` like its three call sites (that last one is the item I offered on 08-26 and held back — a hard build error felt different from a warning). With it, `run_tests.py` is **13/13, 95 tests**, bench targets unaffected. Entirely yours to redo differently; close it and I will follow whatever lands.

A footnote on your ratchet, because it cost me a wrong patch first: while a flavour fails to build no warnings are collected for it, so `test_warning_baseline_is_not_stale` reports its baseline entries as retired. The two `[emu]` lines read as stale for exactly that reason and come back the moment emu compiles — `WARNING_BASELINE` is untouched in that PR.

#### Two merge results, read rather than assumed

**Code fact:** your `calbackup` at `0xFFD000` and this branch's `fwcache` at `0xD00000` compose into a still strictly ascending, non-overlapping table — the self-check passes for the composition, not merely for each half. **Code fact:** the CDC RX reroute wraps the single `shell_feed()` call in the USB branch, and none of `main`'s new shell work lands inside it. Host side: `test-fw-loader` and `test-config-persist` 22/22, `guest-coldtrace` links at 606 364 B.

And your work travelling the other way, with thanks: the 2C23T port picked up `0b7f9d0` and `479d120` on 2026-08-31 — TMR13 armed at a measured centering code, plus an explicit CS-high settle with a WEL read-back. **The 2544 is yours, one unit and one range**, labelled as yours at the definition; unit #2 has not measured its own code yet.

#### State of this PR

No bench runs this week, so nothing re-measured and nothing withdrawn. The one item still labelled untested is the same one: the drain's owed-count fix is host-tested and code-reviewed but never bench-exercised, since it needs a host writing in non-multiples of 64 and neither of our tools does. Still not claiming it.

Two things open from 08-26 whenever suits you, no urgency: whether the reserved 201+ block for unit #2's writeups works for you or you would rather a different scheme, and whether anything else stands between this and merge now that the feed, the tSHSL fix and the RAM resolution are all in. Thanks again for the review — this has been an unusually pleasant PR to keep alive.

DavidClawson:
Thanks — and the `.PHONY`-at-the-cause fix (`fcb53da`) is the right instinct; giving `test-fw-loader` its own line so the shared one stays byte-identical with mine should stop that papercut recurring for both of us. I've replayed your composition check on my side too: `calbackup` at `0xFFD000` and `fwcache` at `0xD00000` do compose into a strictly-ascending, non-overlapping region table, and the self-check passes on the composition.

On the main build break you flagged: fixed now — #31 landed the compile guard and I cleaned up a RAM overflow it exposed (`545a9d8`), so `main` links across all flavours again. Your instinct to keep it out of this PR's diff was correct.

On the two opens: the reserved 201+ block for unit-#2 writeups works for me — go ahead. And I think the feed, the tSHSL fix and the RAM resolution being in means this is merge-ready from my side; the only thing still carrying the "untested" label is the drain owed-count fix, and I'm content to merge it labelled as host-tested-not-bench-exercised rather than hold the PR for a tool neither of us has. If you agree, I'll do a final review pass and merge. This has been a genuinely pleasant one to keep alive — thank you for the care.

Stlkv:
@DavidClawson — thank you, and yes to the label: merging the owed-count fix as host-tested-not-bench-exercised is the right call, I am happy with that.

One catch before you press the button, though, and it is mine this time. Merged with main at 545a9d8, this branch was 568 B over the RAM ceiling on make guest and make — measured; main alone links with 648 B to spare. Your bench flavour guest-coldtrace links fine, so a review pass would not have shown it, exactly like the edgecap case.

What I did about it
The loader no longer owns a buffer (dfede95). I borrowed your idiom from 545a9d8: it now uses shell_bus_scratch, the arena spi3 read/spi3 frame share, on the same grounds — every loader entry point runs on the shell task, and nothing can interleave with a transfer (RX is routed to the loader) or an install (interrupts off). The shell task lends the arena through fw_loader_attach_scratch(), so the loader stays ignorant of USB and the host test lends its own array; begin()/install_slot() refuse with a new ERR_NO_BUFFER until something is lent (host test added). The installer still works each 2 KB page in four 512 B passes, byte-for-byte the bench-proven mechanism. Measured: 568 → 48 B over.
The last 48 B are yours to decide — a45c56f is a proposal, not a decision: configTOTAL_HEAP_SIZE 32 → 31 KB. The loader's remaining state is 43 B in total and its 680 B installer has to live in RAM, so I could not find them on my side without cutting into a proven installer for a few dozen bytes. The heap's low-water mark, measured tonight on the new build, is 13 168 B free; on 08-26 it was 12 128 B in X-Y. One build, shell-only session — a reading, not a law — and the comment at the define says to re-measure before taking any more. Drop that commit if you would rather find the room elsewhere; nothing depends on it. With it, every flavour links (guest headroom from the map: 1 044 B), run_tests.py is 13/13.
Bench, unit #2, tonight (EXP-204)
Control first: your loader on the port side staged the new 607 012 B image (STAGED, crc F7D8EDFD, 30.9 s) and fwswap b booted it. Then on the new build: spi3 read gives a live baseline, fwload of the same image through the new loader → STAGED 607012/607012 crc=F7D8EDFD err=none (64.9 s), spi3 read still live afterwards, heap low-water unchanged, and fwswap a — the new installer reading through the borrowed pointer — put the port back with the cache table intact. Not run: the torn-transfer paths, spi3 frame, and fwapply itself (same installer, other slot). Also unchanged and worth knowing: this firmware streams at 9 KB/s where the port does 19 — not new (67 s on 08-25 too), not isolated.

So from my side: if the 1 KB is fine by you, merge away; if not, say where you would rather find it and I will follow. Either way, thank you for how this one has gone.

DavidClawson:
@Stlkv — **yes to `a45c56f`, take the kilobyte.** Merging now. Sorry for the wait on a one-word answer.

I reproduced the constraint here rather than taking it on report, and it holds on today's `main`:

| `configTOTAL_HEAP_SIZE` | `make guest` |
|---|---|
| 32 KB (before your commit) | fails to link, `region RAM overflowed by 40 bytes` |
| 31 KB (`a45c56f`) | links, **2,088 bytes** spare |

You measured 48 bytes and I get 40, which is just the couple of commits `main` has moved since your merge base. Same conclusion.

**Why the heap and not 40 bytes of static allocation somewhere else.** Your heap low-water of 13,168 B free out of 31,744 says about a third of that arena is never touched at peak. Handing back 1 KB leaves roughly 12 KB of *measured* margin. The alternative is spending real time hunting 40 bytes to preserve a resource we have evidence we are not using, which is the worse trade. The thing to watch is the low-water mark itself: if a future change pulls it down toward 1 KB, this decision needs revisiting, and the low-water number is the signal, not the link succeeding.

**Verified the merge result before pressing the button**, since the last two RAM problems both linked fine in the bench flavour and broke elsewhere:

- `guest`, `guest-coldtrace` and `emu` all build clean from the merged tree.
- All 13 host test binaries pass, `test-fw-loader` included.

**Merging with the label we agreed:** host-tested, bench-exercised on unit #2, and explicitly **not** exercised on torn transfers, `spi3 frame`, or `fwapply`. That is on the record so nobody reads the merge as more validation than it carries. Your EXP-204 round trip, control first, is the reason I'm comfortable at all: a 607,012 B control stage through the port-side loader, then the same image through the new path with a CRC match and `spi3 read` live on both sides, is a real acceptance test rather than a smoke test.

I'll put it on unit #1 next bench session and cover the three gaps. If it survives a torn transfer on a second unit, the label can come off.

Thank you for the whole arc of this one: finding your own RAM overflow, reporting it unprompted, and fixing it by borrowing `shell_bus_scratch` instead of growing the image. That last part is the good version of the fix.

PR #31, opened 2026-09-02
fpga: build the EXP-37 edge-capture rig only where its bit-bang half exists

Stlkv:
`make guest`, `make` (app) and `make emu` do not compile on `main` at `0b7f9d0`. **Measured**, clean worktree of `origin/main` with nothing of mine in it, `arm-none-eabi-gcc 14.2.Rel1`:

```
src/drivers/fpga.c:5816:18: error: 'BB_SCK' undeclared (first use in this function)
src/drivers/fpga.c:5816:27: error: 'BB_CS' undeclared (first use in this function)
src/drivers/fpga.c:5817:18: error: 'BB_MOSI' undeclared (first use in this function)
src/drivers/fpga.c:5823:26: error: 'BB_MISO' undeclared (first use in this function)
src/drivers/fpga.c:5830:9:  error: implicit declaration of function 'bb_cmd16'
make: *** [build/fpga.o] Error 1
```

**Code fact:** `fpga_edgecap_15()` (yours, `e803915`) is compiled unconditionally, while `BB_SCK`/`BB_MISO`/`BB_MOSI`/`BB_CS` and `bb_cmd16()` all live under `#if FPGA_CONFIG_B`. `guest-coldtrace` and `guest-bringup-bb` are unaffected — **measured**, both still build at `0b7f9d0` — which is very likely why the bench never saw it, since those are the two flavours we each flash. That last clause is an inference, not something I measured about your bench.

### What this changes

1. `fpga_edgecap_15()` and its prototype get the same `#if FPGA_CONFIG_B` as the transport they use. The rig is an AF-vs-bit-bang comparison, so it can only mean anything on a `FPGA_CONFIG_B` image anyway.
2. The `spi3 edgecap` shell command stays in the table on **every** build and answers with the flavour to flash, rather than disappearing from `help` and leaving the operator to guess why. Judgement call — say the word if you would rather it were absent.
3. `spi3_xfer2_gapless()` is gated on `FPGA_HW_PRELUDE_GAPLESS`, which is what its three call sites already use. Without that it is an unused-function warning in every default build and your own ratchet in `scripts/test_firmware_build.py` fails app, emu and guest on it. This is the item I offered to push on 2026-08-26 in #29 and did not, because it was your file; the hard build error above changed my mind about waiting. The knob's `#ifndef` default moves up a few hundred lines to sit above the helper, so the guard can read it.

`WARNING_BASELINE` is deliberately **untouched**. The two `[emu]` entries it lists (`vMeterAutoselectTask`, `vUsbDebugTask`) look stale while emu fails to build — no build, no warnings collected, and `test_warning_baseline_is_not_stale` then reports them as retired. With emu compiling again both warnings come back, exactly as the baseline says. I pruned them, saw them return, and put them back; worth knowing that the staleness half of the ratchet reads as a false positive whenever a flavour is broken.

### Verified

| claim | evidence |
|---|---|
| `main` at `0b7f9d0` fails app/emu/guest | measured, clean `origin/main` worktree, output above |
| this branch builds all five gate flavours | measured, `scripts/run_tests.py` 13/13 suites, 95 tests, toolchain present |
| bench targets still build | measured: `guest-coldtrace`, `guest-bringup-bb`, `guest-configA-gapless`, zero errors each |
| `spi3 edgecap` still works where it matters | **not measured on hardware** — it is compiled into `guest-bringup-bb` as before, but I have not re-run the rig with a logic analyser on this build |

No behaviour change is intended for any image: everything moved is either bench-only code or a preprocessor guard matching an existing call-site guard. If you would rather fix this differently in your own file, close this and I will follow whatever lands.

### DIFF
--- a/firmware/src/drivers/fpga.c
+++ b/firmware/src/drivers/fpga.c
@@ -390,6 +390,22 @@ static uint8_t spi3_xfer(uint8_t tx_byte)
     return (uint8_t)FPGA_SPI->dt;
 }
 
+/* [H-gapless / post-H6 suspect (a)] Clock each mode-0 prelude frame's two bytes
+ * GAPLESSLY via spi3_xfer2_gapless() instead of the stalling spi3_xfer(), so SCK
+ * does not idle between the command byte and its 0x00 param inside the CS-LOW
+ * frame — matching the continuous clocking of both working transports (bit-bang
+ * + stock). This is the LAST untested firmware axis after H1-H6: register/byte/
+ * framing are all excluded, so the only firmware-reachable difference left is the
+ * intra-frame cadence our polled peripheral introduces. 0 = shipping path
+ * (spi3_xfer, stalling). Test: `make guest-configA-gapless`. Mode-0 only.
+ *
+ * The default lives here, above spi3_xfer2_gapless(), so the helper can be
+ * compiled only for the builds that call it. */
+#ifndef FPGA_HW_PRELUDE_GAPLESS
+#define FPGA_HW_PRELUDE_GAPLESS 0
+#endif
+
+#if FPGA_HW_PRELUDE_GAPLESS
 /* [H-gapless] Send two bytes back-to-back with NO inter-byte stall — the
  * double-buffered idiom of spi3_pump_h2_record(), for a 2-byte prelude frame.
  *
@@ -427,6 +443,7 @@ static void spi3_xfer2_gapless(uint8_t b0, uint8_t b1, uint8_t *r0, uint8_t *r1)
     if (r0) *r0 = rx0;
     if (r1) *r1 = rx1;
 }
+#endif /* FPGA_HW_PRELUDE_GAPLESS */
 
 static void fpga_h2_record_body_rx(uint8_t rx_byte)
 {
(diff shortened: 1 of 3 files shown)


DavidClawson:
Merged — thank you, and apologies for the breakage, which was mine (`e803915` compiled `fpga_edgecap_15()` unconditionally while its bit-bang symbols sit under `#if FPGA_CONFIG_B`). Your diagnosis of *why our bench never saw it* is exactly right: `guest-coldtrace*` builds clean, and that's what we flash, so the failure hid in the flavours we don't.

Confirmed here across `guest`, `emu`, `guest-coldtrace-ch2` and `guest-bringup-bb` — all link after the merge. Keeping `spi3 edgecap` in the shell table on every build with a "flash `guest-bringup-bb`" message is the right call; a command that silently vanishes from `help` is worse than one that tells you which image it needs.

One follow-on worth noting: the merge exposed a *second* latent bug of mine underneath yours — my `spi3 frame` command (landed just before your PR) added two 1 KB static buffers that overflowed guest RAM by 368 bytes, which never surfaced because the compile break upstream meant guest never reached the linker at all. Fixed in `545a9d8` by sharing one scratch buffer with `spi3 read`. So your PR didn't just fix the compile error, it unmasked a RAM regression — thanks twice over. And the note about `WARNING_BASELINE` reading stale while a flavour is broken is a sharp catch; that's a real gap in the ratchet.

Issue #15, opened 2026-06-06
[Feature]: Get multimeter working


Komzpa:
### What kind of request is this?

New measurement or analysis (FFT variant, math channel, protocol decoder, etc.)

### What are you trying to do?

Hi. I bought this device primarily as "cool multimeter" and will be happy is multimeter works in firmware.

I tried to get Codex on it, so #13 has my progress on it in the last couple of days. I plugged in bench PSU and debugged voltmeter on range of 32..0.8V, and short-circuit. For now values between 0 and 0.8 are --- as they are getting reported in some different shape. I'm out of tokens after 48h of letting it debug it so here's the current progress :)

While at it Codex implemented some more debug protocol to drive fpga via usb and check its hypotheses and fixed more of the screen flicker (not finally, need double buffering for proper fix).

Let me know if #13 has good enough shape to move forward or what I should clean up there.

### Would you be willing to help build this?

Yes — I can write code and test on hardware

DavidClawson:
@Komzpa — this has been sitting for two months and that's squarely on me. Worse, the thing you spent 48 hours of Codex tokens deriving was already solved in the tree, and nobody told you. Let me fix that properly.

**The meter data path works, and has since early April.** DCV and the full resistance range read accurately within a few percent on bench unit #1. The decoder is `firmware/src/drivers/meter_data.c`; the full trace from USART frame to displayed value is annotated in `reverse_engineering/analysis_v120/meter_math_pipeline_annotated.c`. Polling is decoupled from the UI in `fpga_meter_poll_task` (`fpga.c`, ~4 Hz in meter mode).

**Your 0–0.8 V "reported in some different shape" is almost certainly the frame[6] rotation**, which is the single nastiest thing about this device and cost us a while too. The meter IC rotates through *several different `frame[6]` variants for the same measurement*, and each one claims a different decimal-point interpretation of the **identical** raw BCD. So the same resistor would alternate between `9.821 kOhm` ✓ and `98.24 kOhm` ✗ depending on which variant arrived first. The fix is to stop trusting the per-frame dp hint and decide at the **band** level — upper nibble 0 → low-Ω, upper nibble 4 → kΩ. See `meter_data.c:146-178`. If a value is landing as `---`, look at which `f6` bytes you're actually receiving before trusting anything downstream; there's a diagnostic history buffer (`meter_f6_history`) already wired up for exactly that.

[...]

PR #33, opened 2026-09-12
meter: the SoC listens to AA 55 — turn the header on and send it the words it obeys

Stlkv:
@DavidClawson — PR (1) from #15, on top of `7175905` (your keepalive gate and `meter_build_tx_frame()`). Three commits: table + header, the transition fixes the bench forced, and EXP-205.

### What changes

1. **One measured word per submode** (`stock_meter_word_low_for_submode[]` in `fpga_meter_plan.c`). `stock_mode` stays as the formatter-family index for `meter_data.c` and the mux projection; it no longer picks the wire word. `fpga_meter_stock_cmd_low_for_mode()` and the apply-word function are gone — the four "apply pairs" are stock's AC/DC toggle, and the second word of each is now the primary selector of ACV, Diode, AC mA, AC A.

   | submode | word | was | | submode | word | was |
   |---|---|---|---|---|---|---|
   | 0 DCV | `050C` | `0514` (Auto) | | 6 Resistance | `050B` | `050A` |
   | 1 ACV | `050D` | `050C` | | 7 Continuity | `0517` | `0511` |
   | 2 DC mA | `0511` | `0517` | | 8 Diode | `050E` | `0510` |
   | 3 DC A | `0510` | `0517` | | 9 Capacitance | `050A` | `0512` |
   | 4 AC mA | `0516` | `050B` | | 10 Temperature | `0512` | `0512` |
   | 5 AC A | `0515` | `050B` | | | | |

   The `14 0c 17 0b 0a 12 11 10` table at `0x080BB3FC` is real (your literal test still pins it), but it is stock's **menu order**, not a mode index.

2. **Header default ON.** `meter hdr off` stays as the negative control.

3. **OBEY is decided by the caller.** Obeyed: the transition's selector (`fpga_wire_send_word`) and the operator's `fpga_send_cmd` / `usart tx` / `fpga wire words`. Everything that has gone out `00 00` since April keeps `00 00`, bit-identical: wake preamble, `fpga_init` and poll-task activation blocks, siggen entry, every scope/config sequence via `fpga_timed_send_cmd`, and the shell's batch replays (`fpga wire entry/scope`, `stock diag bridge`, `spi3 scopetest`). This is the EXP-26 audit: `0x050A` (your "PC7-low probe tail") is the SoC's **Capacitance** selector and `0x0514` ("variant setup") is **Auto**; both sit a few ms after the selector in the mode-entry sequences and, obeyed, would have undone every transition.

4. Header and table are one commit on purpose: under `00 00` a wrong word is discarded; under `AA 55` it is obeyed. Header without table is worse than either alone.

Also: `fpga.h` labels for `0x12/0x13/0x14` now read Temperature/LIVE/Auto; `docs/stock_vs_openscope.md` meter rows say "selected, decoder pending"; `scripts/validate_dmm_goal.py` pinned the old table by regex, anchors updated — your file, redo as you like.

### Bench — EXP-205, unit #2 (`docs/experiments/2026-09-13-205-word-table-and-header-on-unit2.md`)

1. `mode meter N` for N = 0..10: `echo_valid` +1 each, `echo_bad` 0 (24 echoes / 0 bad in the session), frame signature of function N. `meter hdr off`: five sends, no echo, SoC stays in its power-on state.
2. 30 s of keepalive (129 frames): no echo growth, ~6.5 data frames/s, **no relay actuation** (operator, relay pose read back).
3. Loads, hand-decoded: cell 3.862 V; 2.2 kΩ → `2168`; 10 kΩ → `9.924` k; 300 kΩ → `2978` under `f7=0x24` (297.8 k); short → ` 0.17` Ω in Ω and Cont; 10 nF → `9.623` nF; 10 µF → `9.697` µF. Your decoder gets DCV and the two kΩ readings; the rest come back `---`/`ERR`/×100 — PR (2), with these as fixtures.
4. SPI3 15.3 → 14.2 reads/s across scope→meter→scope, 0 timeouts.

### Two defects the bench found, fixed in the second commit

1. **The transition's selector was never accepted.** `AA 55 05 0B` left the wire, no echo came, and **no data frame arrived during the transition** (`data=558..558`); the next frame was the SoC's power-on "Auto" text. `fpga_meter_reset_transport()` drops PC11 and the SoC **restarts**; it is deaf for **1380 ms** (24/24 transitions, a constant on this unit), and the selector 20 ms later is lost. By hand a second later the same word echoed at once. Fix: `fpga_meter_wait_for_soc()` (first frame after PC11, ≤3 s, unobeyed keepalive meanwhile, +200 ms) and `fpga_send_selector_confirmed()` (re-send until echoed, ≤5 × 300 ms). `meter hdr` prints `transition: wake_ms= tries= ok= | unconfirmed= wake_timeouts=`.
2. **One scope→meter entry in three came up with `CTRL1=0x20A0` — UEN set, TE=0 RE=0**: TX climbing into a disabled transmitter, RX frozen. Scope entry parks USART2 with `ctrl1=0`; the reset reads that, quiets 20 ms, restores `ctrl1|UEN|RDBFIEN`; when the display task's `fpga_set_meter_mux(true)` lands inside those 20 ms, the restore overwrites it. The reset now sets TE/RE and the IRQ itself. After: 8/8 first-try, `0x202C` every time.

Something the clicks said: DC mA → DC A produced two clicks with our relay pose unchanged (read back, byte-identical); Cap → Temp none. The SoC has relays of its own and moves them only for an obeyed word that needs it — consistent with the #15 logger (stock's MCU moves no frontend pin across 40 function changes).

### Weaknesses

1. Currents and ACV: word accepted and frame signature seen, no value checked (no source). Temperature: internal sensor only (`  28`).
2. Loads are ±5 % parts and an unlabelled cell; nothing here is calibrated.
3. The 1380 ms wake is one unit's constant; the 3 s timeout has 2.2× margin here only.
4. The SoC's reaction to an obeyed non-`0x05` frame is unmeasured — hence every such sender stays unobeyed rather than "probably fine". On `guest-coldtrace-meter` nothing is transmitted in scope mode, so the scope-side control was vacuous.
5. Legacy sequences are kept as unobeyed traffic rather than deleted: conservative by choice; deletion is a one-liner if you agree that `0x0508/0x0509` and the `00 2C` prefix are display bytes (our reading of the stock binary, hypothesis-grade).

### Three questions, not changes

1. Stock never moves PC11 between functions (#15 logger), so the restart — and 1.4 s per function change — is ours. Your `verify_meter_transition_usart_gate` pins the drop; I left it. Want it gone?
2. EXP-26's "transmit stopped → 0 frames/s" arm was `mode scope`, which also drops PC11. "Answers traffic" may be partly "answers when powered" — worth one re-run with PC11 held high.
3. Your EXP-27 short: `04 E0 1B 4A` has the DP bit in the pre-lookup nibble of digit 2 (`1B` → nibble `0x1A`), so the text is ` 0.14` = **0.14 Ω**, not 0.014 — `0x10` on the leading digit is the blank glyph after lookup. Our leads read ` 0.17` with the same encoding. Prediction: `meter dump` on that frame shows `dbg_nibbles[2] & 0x10`.

### Tests and builds

`make -C firmware test-meter` and every other `test-*` target green; `run_tests.py` 13/13 suites, 95 tests, ratchet clean; `test_dmm_goal_validation.py` 11 OK. All five flavours link: `guest` 614 044 text / .bss +8 vs `7175905` (four counters), `guest-coldtrace` 607 848, `guest-coldtrace-meter` 608 584, `emu` 269 560, `app` 614 148.

Thank you for the replication, for catching the second frame builder, and for the poll gate — without it this would have been a relay-clicking bug report.

### DIFF
--- a/docs/experiments/2026-09-13-205-word-table-and-header-on-unit2.md
+++ b/docs/experiments/2026-09-13-205-word-table-and-header-on-unit2.md
@@ -1,0 +1,210 @@
+# EXP-205 — The corrected word table under AA 55, on unit #2
+
+- **Date:** 2026-09-13
+- **Unit:** bench unit #2 (Stlkv)
+- **Build:** `make guest-coldtrace-meter`, branch `meter-word-table` (this PR), staged into
+  slot B over the port firmware's CDC loader and installed with `fwswap b`
+- **Status:** **CONFIRMED** — with two defects found and fixed on the way, both in this branch
+
+## 1. Problem
+
+PR (1) of issue #15 turns the meter TX header on by default and replaces the word table.
+Nothing on the bench had ever exercised a *UI-driven* transition with the header live:
+EXP-25/26/27 commanded the SoC by hand (`usart tx`) after entering meter mode. Does
+`mode meter <submode>` select the function it names, does the unobeyed keepalive stay
+silent, and does every one of the eleven words get accepted?
+
+## 2. Hypothesis
+
+With the header on and the measured table in place, each `mode meter N` produces exactly
+one accepted echo and the SoC's frame changes to the signature of function N; the 4 Hz
+keepalive produces no echoes and no relay actuation.
+
+- **If true:** `echo_valid` +1 per transition, `echo_bad` 0, no echo growth between
+  transitions, data frames ~6/s, frame signature per function.
+- **If false:** the echo does not move on a transition, or moves without one.
+
+## 3. Procedure
+
+```
+meter hdr                # default must read AA 55
+mode meter [N] 0         # UI transition path (fpga_set_meter_mode -> fpga_apply_meter_transition)
+meter hdr / status       # echo ladder, tx_frames_recent, transition counters
+meter dump               # frame, nibbles, raw_digits
+meter hdr off            # negative control
+```
+
+Frames are hand-decoded from `nibbles=` (pre-lookup): bit 4 of a digit's nibble is the
+seven-segment decimal point, lit on the digit it precedes; the glyph is the other seven bits.
+
+**Preconditions verified by readback:**
+
+| what | expected | measured |
+|---|---|---|
+| header default | `AA 55 (default)` | ✅ |
+| scope mode, 10 s | TX 0, RX 0 on this flavour (`FPGA_USART_SILENT_SCOPE=1`) | TX 0, RX 0 — so the scope-side control is vacuous here, see §6 |
+| keepalive on the wire | `00 00 05 09` | ✅ every frame in `tx_frames_recent` |
+
+## 4. Control
+
+Positive control, recorded first, on the **first** build of this branch (before the fixes
+below): `usart tx 05 0C` by hand in meter mode → `echo_start 0→1, echo_valid 0→1,
+echo_bad 0`, frame changed from the boot-time `E4 2E 63 25 07 00 00` to a DCV frame
+(`f8=0x82`, reading 0.0003 V). The header and the RX echo path work on this build.
+
+Negative control (final build): `meter hdr off`, then `mode meter 6 0` → the selector was
+sent five times under `00 00`, `echo_valid` unchanged, `confirmed=0`, and the SoC's frame
+stayed the boot-time text — the SoC does not obey `00 00`.
+
+## 5. Results
+
+### 5a. Defect 1 — the transition's selector was never accepted (fixed in this branch)
+
+First build, `mode meter`: TX count climbed, keepalive frames flowed, **`echo_valid` stayed
+0**. `mode meter 6 0`: `tx_control_history` shows `AA 55 05 0B` leaving at tx=401 — the
+selector was transmitted — and the transition history shows `data=558..558`: **not one
+data frame arrived during the transition.** The frame afterwards was the boot-time
+`E4 2E 63 25 07 00 00` (the SoC's power-on "Auto" text), not the DCV frame the SoC had been
+sending a second earlier.
+
+Reading: `fpga_meter_reset_transport()` drops PC11, and the SoC **restarts** — every
+transition lands it back in its power-on state, and it is deaf while it boots. The selector
+20 ms later goes into that silence. The same word typed by hand a second later was echoed
+at once.
+
+Fix: `fpga_meter_wait_for_soc()` waits for the first data frame after PC11 comes back (up
+to 3 s, feeding unobeyed keepalive traffic every 250 ms), then +200 ms; the selector is
+sent by `fpga_send_selector_confirmed()`, which re-sends until its echo arrives (≤5 tries,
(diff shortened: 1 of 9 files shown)


DavidClawson:
Merged as `12c038c`. Label on the record: **bench-exercised on unit #2 only**, host-tested here — 4 of 4 build flavours, 15 of 15 test targets, 13 of 13 suites. Unit #1 has not run it yet.

**Your root cause is better than the one we had, and it matters.** We had established *that* the assignment was wrong for ten of eleven submodes. You found *why*: the eight bytes are real, but they are stock's **menu order**, and we were indexing them by a mode number. That reframes it from "somebody mis-typed a table" into a specific mechanical error anyone could make again, which is the version worth having written down. It also explains the shared-slot fallbacks without appeal to hardware: the four words we lacked are exactly the four that forced the sharing.

**The transition finding is the one I did not see coming.** Every mode transition this firmware has ever made selected nothing, because the SoC restarts and is deaf for ~1.4 s and our selector went out 20 ms in. That is the sixth instance this week of the same shape: a control that could not have had an effect, so nothing it did was ever wrong, so nothing looked broken. Waiting for the first frame and then confirming the selector against its echo is the right fix — it stops the transition taking anything on trust. 24 of 24 at 1380 ms is a real number, not an impression.

**A cross-check you should know about, because neither of us set it up.** Before your PR was visible I wrote a host test from the twelve-word map in your issue comment — the map as the reference, our mapping as the thing under test, expected to fail. It landed on `main` red a few hours before your branch arrived. It passes against your code, unmodified in its reference half. Independent transcription, independent implementation, same answer. I trust the table considerably more for that than I would from either half alone.

Two things I changed while merging, both ours not yours:

- Five conflicts in `usb_debug.c` between your unobeyed senders and a shell sweep that landed here last night. They were complementary, so I took both: `acqtest_send()` now sends unobeyed *and* counts enqueue failures.
- The test above called `fpga_meter_stock_cmd_low_for_mode()`, which your PR deletes. The deletion is the fix, not a regression — that accessor *is* the menu-order indexing bug — so I moved the check onto the new interface. Its converse survives and still passes: every word we can emit is one stock is known to send, which is what says the RE was always sound and only the assignment was wrong.

Unrelated and pre-existing, so you do not need to look at it: `scripts/validate_dmm_goal.py` fails on `main` too, a forbidden-phrase search matching "order of magnitude" in scope calibration comments. Our gate, our problem.

Next here: unit #1 on the bench, and the eight words that have still never been on our wire — they need a capacitor, a diode and a thermocouple, so that is a session with parts on the desk rather than a code change.

Of your remaining two, **I would take (3), the logger, first.** It is a general instrument for this project and not a meter tool; the decoder is downstream of it in a way that the logger is not downstream of anything. Thank you for this one.

Issue #10, opened 2026-04-14
[Feature]: PC remote view/control software?


johnathanhawn:
### What kind of request is this?

Other

### What are you trying to do?

Having realtime viewing would be amazing on a bigger screen also remote controls 

### Would you be willing to help build this?

Yes — I can write code and test on hardware

DavidClawson:
@johnathanhawn — 119 days without a reply to someone who offered to help build the thing. That's not good, sorry.

**To be clear up front: this is on the roadmap and we want it.** Realtime viewing on a bigger screen with remote control is a feature this device should have, and it's not going to be quietly dropped. What I owe you is an honest picture of the sequencing rather than a vague "someday."

**What has to land first.** Remote *waveform* viewing runs into the same wall as everything else here: **there is no live scope data to stream yet.** The FPGA won't accept its configuration from our firmware, so the scope captures nothing. That's #18, it is the project's critical path, and it is actively being worked — there was bench and analysis work on it today. Building a waveform viewer ahead of that would mean building a viewer for a signal that doesn't exist.

Second, smaller: **the app's USB CDC channel doesn't currently enumerate reliably.** The bootloader's USB is solid, but the application's serial channel is a known open problem, so the obvious transport needs fixing before anything streams over it.

**But there's a real slice available right now, if you want it.** The **multimeter data path works** — readings flow over USART2 from the meter IC, get decoded in `firmware/src/drivers/meter_data.c`, and reach the UI at ~4 Hz. A host-side tool that streams and logs *meter* readings, plots them, and drives mode/range changes remotely is buildable today against working data.

That's not a consolation prize. Long-duration logging is exactly what a handheld's screen is bad at, it's genuinely useful on its own, and it establishes the host protocol and transport that the waveform viewer will reuse when #18 clears. It's the natural first slice of this feature rather than a detour around it.

Related and worth a look: **#8** is the platform/distribution roadmap, including a browser-based flasher. If host-side tooling is where your interest lies, that thread and this one converge.

**Two things to be straight with you about, since you offered to build rather than just to use.**

First, **I'm not going to spearhead this myself until the FPGA/scope link is solved.** That's where my time is going, and everything else is deliberately queued behind it — partly for focus, partly because until the scope captures we don't really know what the host-side picture should look like. So if you build this, you'd be leading it rather than joining something already in motion.

[...]

PR #34, opened 2026-09-12
scripts: stock-caplog — the logger that measured the meter word map, packaged

Stlkv:
@DavidClawson — PR (3) from #15, independent of #33 (branches from `7175905`). You asked for this one early; here it is.

### What it is

Two six-byte hooks in the user's own stock V1.2.0 APP image (the dvom TX task right after a frame to the meter SoC is built, and the SysTick handler) jump into a 692-byte code cave appended at `0x080BE700`. The cave logs every USART2 frame to the SoC (ring of 32) and every GPIO ODR/CFG change on ports A–E (ring of 64) plus a per-pin toggle counter, into SRAM `0x20037000..0x20037700`. The block survives a MENU+Power round trip and is read back over a debug shell. No logic analyzer, no case opening.

- `scripts/stock_caplog/build_caplog_image.py` — compiles the cave (no `.data`/`.bss`, no libc, integer registers only; the linker script asserts the first two), verifies the stock image by sha256 and the six original bytes at each hook, changes exactly 12 bytes, appends the cave. Refuses anything but V1.2.0. Nothing stock is redistributed.
- `scripts/stock_caplog/caplog_dump.py` — reads the block and decodes it; `--decode` re-renders a saved one.
- `scripts/stock_caplog/samples/` — the three unit #2 dumps of 2026-09-07 that the word table in #33 rests on.
- `scripts/test_stock_caplog.py` (13 tests, registered in `run_tests.py`) and `docs/stock_caplog.md`.

### What the samples say (measured, unit #2)

`run2`: stock's meter menu walked twice → `0C 17 0B 0A 12 11 10 14`, i.e. menu order, Capacitance = `0x0A`. `run3`: the AC/DC toggle inside a function is its own word — `0C→0D`, `17→0E`, `11→16`, `10→15`. `run1`: the walk that settled `0x0A` (capacitance) vs `0x0B` (resistance). Across all three, the only GPIO traffic is the key-matrix scan (63 toggles per pin in run3) and a 1 ms pulse train on PB12 (820): none of the frontend pins this firmware drives per submode moved. The tests pin both — the words against `fpga_meter_plan.c`'s per-submode table (that test skips until #33 lands) and the frontend against the toggle counters.

### Verified

- Build reproduces the image flashed on 2026-09-07 byte for byte with Arm GNU 14.2.Rel1 (sha256 `3ce0edcb…`).
- `run_tests.py` 14/14 suites, 108 tests; the three skips without the stock archive / toolchain are reported as skips.
- Live read exercised on unit #2 today through both shells (this firmware's `mem read <addr> 64` and the 2C23T port's `mem <addr> 256`): 7 reads, 1792 bytes, address-checked.

### One limit I found while packaging — please read

**This firmware cannot read the block back.** Its `.bss` covers `0x20037000` (`_ebss` = `0x200375D4` in the app flavour, higher in `guest`) and startup zeroes it — measured today: after booting `guest-coldtrace-meter` the block reads all zeros. The stack sits above `.bss` up to `0x20037FE0`, so there is no free window to move it into. The reader has to leave that SRAM alone; the 2C23T port does (RAM ends `0x20036078`), which is how the samples were taken. For units with your HID bootloader the candidate is `caplog_dump.py --hid` via `hid_flash.read_memory` — **untested**, and whether the bootloader's own startup touches that SRAM is the first thing to check. `build_stock_hybrid_image.py` also pins the clean stock sha256, so on those units it would need a flag to take a caplog image knowingly. I did not touch it.

### Weaknesses

1. V1.2.0 only — hook addresses and displaced instructions are facts about that image; a new stock version needs the two sites found again.
2. SRAM survival is measured with the factory IAP bootloader on one unit (three round trips). Not on your bootloader/dispatcher path.
3. The SysTick hook does five GPIO reads and compares per tick; stock ran normally with it, but its effect on stock's timing was not measured.
4. `--hid` is code I could not run.

Thank you for wanting this one early — it is the instrument behind #33, and it should answer the next "what does stock actually do" without anyone's Saleae.


DavidClawson:
Merged as `7d45cc0`. Thank you for promoting this ahead of the decoder — it is the general instrument I hoped it was.

**Verified independently of the script's own checks**, against the archive image on this machine:

- sha256 `a17c5c35…0c760`, length `0xB7680` — matches `CLEAN_SHA256` / `CLEAN_LEN`.
- Both hook sites disassemble to exactly the six displaced bytes: `0x0803E44C` = `movs r0,#10; bl 0x8041390`, `0x0802A994` = `push {r7,lr}; mov.w r0,#16`.
- Vector 15 of the stock image is `0x0802A995`, so site 2 is the real SysTick handler, not a look-alike.
- Built the image: 752,052 B, 12 bytes changed inside the original, 820 appended. The patched hooks decode as `b.w 0x80be700; nop` and `b.w 0x80be712; nop`, and the cave's literal pool carries `0x08041391` / `0x0803E453` / `0x0802A99B` — the thumb-bit-correct return targets.
- `test_stock_caplog` 13/13.

Label: **not flashed here; bench-exercised on unit #2 only.** I have no reason to expect a difference on unit #1 but will say so only after it has run.

One design note for the record, no action needed: `caplog_tick` runs *before* the handler raises BASEPRI, so it executes at SysTick priority with the task-level `caplog_tx` able to be interrupted mid-record. The two never share an index (separate `tx_count` / `gpio_count`), and the worst case in `caplog_init` racing itself is a doubly-zeroed block that ends consistent. Fine as is.

PR #35, opened 2026-09-12
meter_data: the frame is seven-segment text — read its point, its blanks and its bands

Stlkv:
@DavidClawson — PR (2) from #15. Branches from `7175905`, independent of #33 and #34 in code; the fixtures come from unit #2 (EXP-205, 2026-09-07) and from your EXP-27 frames.

### What changes in `meter_data.c`

1. **The decimal point is in the frame.** Bit 4 of a digit's nibble (before `bcd_lookup`, which masks it) is the seven-segment DP, lit on the digit it precedes. Exactly one lit DP sets `decimal_pos` for every submode except DCV, whose point rides your stock range classes and the +10000 extension and is untouched. Your existing mains fixture already carries it (`… ED F7 07` → point on digit 3 → 226.6 V in ACV).
2. **A leading blank glyph is a leading space.** `" 0.17"`, `"  28"`: those digits are zeros for the value and nothing for the display. A frame blank all the way through is still the blank special.
3. **The other spelling of OL.** `" 0L "` — blank, 0, L, blank — is what resistance shows on open probes, continuity above its threshold and diode above its range. It was `ERR`.
4. **Resistance / continuity bands, without a coefficient.** `frame[7] & 0x04` → digits are hundreds of ohms (`"2978"` → 297.8 kΩ); a point in the frame → the text as shown, kΩ when `frame[6]`'s upper nibble is 4 (`"9.924"`), else Ω (`" 0.17"`); no point → integer ohms, shown as `X.XXX kOhm` in the kΩ regime (`"2168"` → 2.168 kOhm, your existing rule) else `Ohm`. The `METER_REJECT_UNRESOLVED_CALIBRATION` path for `frame[6]` upper nibble 0 is gone: the 0.0304 factor you withdrew was the auto-mode artefact, and the low band reads as text on both units.
5. **Capacitance describes itself.** `frame[7] == 0x10` marks the function; the unit is `frame[6]`'s upper nibble, `0x2_` nF and `0x1_` µF; the point is in the digits. Any other nibble is refused (`UNSUPPORTED_EXTENSION`) — mF was never on a bench. Synthetic frames without the `0x10` marker keep the stock formatter's default, so your existing capacitance tests still describe them.
6. Outside DCV, `decimal_pos == 0` now means "the integer shown"; `format_reading()` also strips leading zeros ahead of a point (`"00.17"` → `"0.17"`).

Not changed, on purpose (two things the bench says, both against the current family model — your call):

- **Diode frames carry the voltage marker.** Both diode frames on unit #2 (`" 0L "` above range, `1.346 V` on a cell) have `frame[8] = 0x02`, so submode 8 rejects them as a foreign family. Making the diode family accept the marker touches five of your family-matrix tests; I left it out and the diode still reads `---`.
- **Continuity beeps only on the marker glyphs.** The SoC in continuity sends `" 0.17"` (a number) or `" 0L "`, never the `0x12 0x0A 5` pattern, on unit #2. A number in submode 7 now reads as 0.17 Ohm, `NORMAL`, no beep — as your tests specify.

### Fixtures (all raw 12-byte frames, `test_meter_data.c`)

Unit #2: 10 kΩ `"9.924"`, 2.2 kΩ `"2168"`, short `" 0.17"` in Ω and Cont, 300 kΩ `"2978"`/`f7=0x24`, open Ω `" 0L "`, 10 nF `"9.623"`, 10 µF `"9.697"`, 100 nF, 100 µF, open Cap, the ranging blank frame, an unmeasured unit nibble (refused), Cont on 2.2 kΩ `" 0L "`, temperature `"  28"`, DCV unchanged at `1.4977`.

Unit #1, your EXP-27, reconstructed from the bytes you published (`frame[7] = 0x20` assumed, labelled): 10 kΩ `C4 9F 8A CA`/`0x47` → 9.775 kOhm ✅ your value. Shorted probes `04 E0 1B 4A`: the DP bit is in the pre-lookup nibble of digit 2 (`1B 4A` → `0x1A`), so the text is `" 0.14"` = **0.14 Ω**, not 0.014 — `0x10` on the leading digit is the blank glyph after lookup, not a point. 0.14 Ω is also an ordinary lead resistance where 14 mΩ is not. Prediction you can check: `meter dump` on that frame shows `dbg_nibbles[2] & 0x10`.

### Tests and gates

`test-meter-data` 72/72 (68 + 4 new, one rewritten: the low-band reject became a reading); `test-meter` all green; `test_dmm_goal_validation.py` OK and every static gate in `validate_dmm_goal.py` passes unchanged; `run_tests.py` 13/13. All five flavours link: `guest` 613 880 text / .bss 227 284 (unchanged vs `7175905`), `guest-coldtrace` 607 812, `guest-coldtrace-meter` 608 436, `emu` 269 500, `app` 613 984.

### Bench — EXP-206, unit #2 (`docs/experiments/2026-09-13-206-decoder-reads-the-text-frames-on-unit2.md`)

`guest-coldtrace-meter` built from #33 with this commit on top (the table selects the function, the decoder reads it), loads swapped by hand, three `meter dump` reads each, `reject=0` on every row:

| load | decoder | before |
|---|---|---|
| 10 kΩ | 9.929 kOhm | 9.924 kOhm |
| 2.2 kΩ | 2.169 kOhm | 2.168 kOhm |
| 2.2 kΩ, Cont | OL | `ERR` |
| 300 kΩ | 297.8 kOhm | `---`, reject=4 |
| short, Ω and Cont | 0.16–0.17 Ohm | `ERR` |
| 10 nF | 9.63 nF | 962.3 nF |
| 10 µF | 9.70 uF | 969.7 nF |
| open Ω / Cont | OL / OL | `ERR` |
| internal temperature | 30 C | `---` |

Diode still `---` (`reject=1`), as described above.

### Weaknesses

1. Currents and ACV values are untested; the AC-evidence gate is untouched.
2. mF and the MΩ band were never on the bench; the code refuses the former and has no rule for the latter.
3. Unit #1 fixtures are reconstructions of published bytes, not captured frames.
4. Two ±5 % resistors and unlabelled capacitors; nothing here is calibrated.

### DIFF
--- a/docs/experiments/2026-09-13-206-decoder-reads-the-text-frames-on-unit2.md
+++ b/docs/experiments/2026-09-13-206-decoder-reads-the-text-frames-on-unit2.md
@@ -1,0 +1,89 @@
+# EXP-206 — The decoder reads the SoC's text frames on unit #2
+
+- **Date:** 2026-09-13
+- **Unit:** bench unit #2 (Stlkv)
+- **Build:** `make guest-coldtrace-meter` from the `meter-word-table` branch (#33) with the
+  decoder commit of this PR cherry-picked on top — the word table is what puts the SoC into
+  the function whose frame is being decoded. Installed in slot B over the port's CDC loader.
+- **Status:** **CONFIRMED**
+
+## 1. Problem
+
+PR (2) teaches `meter_data.c` three things measured in EXP-27/EXP-205: the decimal point is
+in the frame, a leading blank is a space, and the resistance/capacitance bands are in
+`frame[7]`/`frame[6]`. The host tests pin the fixtures; does the firmware decoder produce the
+right number on the device, from the live frame, for the loads that EXP-205 hand-decoded?
+
+## 2. Hypothesis
+
+Every load EXP-205 hand-decoded now comes out of `meter dump` as `display=<that value>
+unit=<that unit>` with `reject=0`, three reads apart, stable to the last digit or one.
+
+- **If true:** the table in §5 matches EXP-205 §5d.
+- **If false:** a value differs by a power of ten or a unit, or `reject` is non-zero.
+
+## 3. Procedure
+
+```
+mode meter N 0           # UI transition, echo-confirmed selector (#33)
+meter dump               # x3, 1.2 s apart: display=, unit=, reject=, frame=
+```
+
+Loads changed by the operator between rows; nothing else touched.
+
+**Preconditions verified by readback:** header `AA 55 (default)`; each transition
+`tries=1 ok=1`; frame signatures per function as in EXP-205 §5c.
+
+## 4. Control
+
+Open probes, recorded first, no hands:
+
+| submode | display | reject | frame[2..11] |
+|---|---|---|---|
+| 6 Ω | `OL` | 0 | `04 F0 6B 01 00 24 00 00 01 35` |
+| 7 Cont | `OL` | 0 | `00 E0 7B 01 00 28 00 00 01 35` |
+| 9 Cap | `0.002 nF` | 0 | `E4 FB EB AB 2D 10 00 00 01 37` |
+| 10 Temp | `30 C` | 0 | `00 00 80 EF 0B 00 20 00 01 34` |
+| 8 Diode | `---` | 1 | `00 F0 6B 01 80 00 02 00 01 39` |
+| 0 DCV | `0.0005 V` | 0 | `F6 EB EB CB 07 00 82 00 01 3A` |
+
+Before this PR the same frames gave `ERR` (Ω, Cont), `---` (Temp) and a ×100 capacitance.
+The diode row is the expected negative: its frame carries the voltage marker `frame[8]=0x02`
+and the family model rejects it — left unchanged in this PR, see the PR text.
+
+## 5. Results
+
+| load | submode | decoder, 3 reads | reject | frame[2..11] | EXP-205 hand decode |
+|---|---|---|---|---|---|
+| 10 kΩ | 6 Ω | 9.929 / 9.929 / 9.929 kOhm | 0 | `C4 DF AF CD 4F 20 00 00 01 33` | `9.924` kΩ |
+| 2.2 kΩ | 6 Ω | 2.168 / 2.169 / 2.169 kOhm | 0 | `A4 0D EA C7 4F 20 80 00 01 35` | `2168` Ω |
+| 2.2 kΩ | 7 Cont | OL / OL / OL | 0 | `00 E0 7B 01 00 28 00 00 01 33` | ` 0L ` |
+| 300 kΩ | 6 Ω | 297.8 / 297.8 / 297.8 kOhm | 0 | `A4 CD 8F EA 0F 24 80 00 01 35` | `2978` × 100 Ω |
+| short | 6 Ω | 0.17 / 0.16 / 0.16 Ohm | 0 | `04 E0 1B EA 07 20 00 00 01 35` | ` 0.17` |
+| short | 7 Cont | 0.16 / 0.17 / 0.17 Ohm | 0 | `00 E0 1B 8A 0A 28 00 00 01 35` | ` 0.17` |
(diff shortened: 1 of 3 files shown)


DavidClawson:
Merged as `e9891ae`, on top of #33 so the table and the decoder are on `main` together.

**What I verified here rather than took on report**, on the merge result:

| check | result |
|---|---|
| `make test-meter-data` | 72/72 |
| all 15 C host test targets | 15/15 |
| Python suites | 13/13 |
| `guest` / `guest-coldtrace` / `guest-coldtrace-meter` / `emu` | all link (guest 620,160 B — inside the #29 RAM ceiling) |

Label on the record: **host-tested; bench-exercised on unit #2 only (EXP-206).** Unit #1 has not run the decoder yet — that is my next bench session, and it is where the eight untested words get their capacitor, diode and thermocouple.

Two things I liked in the review and want written down: the mF range is *refused* rather than guessed because it was never on the bench, and the `format_reading` fix (`"00.17"` → `"0.17"`) came with its own fixture. That is the shape of change this repo can trust.

One question, not a blocker: in the `frame[7] & 0x04` upper band the sign is dropped (`kohm = bcd * 0.1`, no `negative` check). I assume resistance cannot be negative on this SoC, but if you have ever seen a `-` glyph in an ohms frame on unit #2 I would like to know.

Stlkv:
@DavidClawson — **no, I have never seen a `-` in an ohms frame on unit #2:** all 16 resistance/continuity rows I have (EXP-205 §5c/§5d, EXP-206 §4/§5) carry `frame[7].0 = 0`, and no minus glyph in the text.

But the band is not where the sign goes missing. `meter_data.c:1371` gates it on the submode, so Ω and Cont leave the parser with `negative = false` whatever that bit holds — upper band, `frame_dp` band and kΩ regime alike:

```c
r->negative = ((submode == 0 || submode == 2 || submode == 3) &&
               (status & (1 << 0)) != 0);
```

That line entered in `626828c` (2026-06-05) and is unchanged at `7175905`, this branch's base.

The bit's meaning is measured, just not in resistance: EXP-205 §5c, submode 2 DC mA, `frame[7] = 0x01`, reading `-00.03 mA` — the only frame in my corpus with bit 0 set, and the only one showing a minus. I never tried to provoke one in Ω; if it is worth closing, I can put a live source across the leads in both polarities next session, and a frame with bit 0 set would come back as a fixture.

Separately, while checking: bit 2 doubles as `is_ac` and the resistance block does not clear it, so the ×100 band prints `flags ac=1` next to `297.8 kOhm`. Diagnostic-only per `meter_data.h:83`, but we can send a one-liner if you want it quiet.

DavidClawson:
@Stlkv — thank you, that closes it, and your reading of the gate is better than my question was.

You are right that the band is not where the sign goes: `meter_data.c:1371` drops it for Ω and Cont by submode before any band is chosen, and that line has been there since `626828c`. I checked all three citations against `main` before replying — the gate, the `is_ac` set at 1369 with no clear in the resistance block, and the "diagnostic only" note at `meter_data.h:83` — and they match.

**Please do not spend a session provoking a minus in Ω.** The one frame with bit 0 set in your corpus is a mA reading with a real sign, which is exactly the measurement the gate encodes. A reversed source across the leads in resistance mode would be testing a condition the instrument has no meaning for, and the gate already refuses to print a sign there. If a bit-0 Ω frame ever turns up by accident, a fixture would be welcome; it is not worth going looking for.

**Yes to the one-liner** clearing `is_ac` in the resistance block. Diagnostic-only or not, a flag line that says `ac=1` beside a kΩ reading is the kind of stable wrong number this repo keeps finding later, and it is cheaper to make it quiet now with the fixture that proves it.

On the label: the decoder is still bench-exercised on unit #2 only. Unit #1's bench time since the merge went to the acquisition path (EXP-50 through 53: the FPGA holds the record for ~189 ms after PC0 and then rolls again, and the trigger sits at mid-record on either polarity), so the decoder plus the eight untested words on unit #1 is still the next meter session, not a done one.

